Issue #301, opened 2024-12-12
Log server instance


lukehinds:
Sometimes its difficult to know if it is the copilot of regular HTTP server, that is logging errors etc.

PR #406, opened 2024-12-18
feat: add custom logger that includes the origin

yrobla:
When running servers, use a custom structlog logger that binds an extra parameter called origin. Can be: generic_server, copilot_proxy

Closes: #301

### DIFF
--- a/src/codegate/logger/logger.py
+++ b/src/codegate/logger/logger.py
@@ -3,4 +3,4 @@ import structlog
 
 class OriginLogger:
     def __init__(self, origin: str):
-        self.logger = structlog.get_logger().bind(origin=origin)
+        self.logger = structlog.get_logger("codegate").bind(origin=origin)
--- a/src/codegate/logger/logger.py
+++ b/src/codegate/logger/logger.py
@@ -1,0 +1,6 @@
+import structlog
+
+
+class OriginLogger:
+    def __init__(self, origin: str):
+        self.logger = structlog.get_logger().bind(origin=origin)
--- a/src/codegate/logger/logger.py
+++ b/src/codegate/logger/logger.py
@@ -1,6 +1,0 @@
-import structlog
-
-
-class OriginLogger:
-    def __init__(self, origin: str):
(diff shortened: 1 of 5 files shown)


yrobla:
sample output:

image

src/codegate/logger/logger.py
```diff
@@ -0,0 +1,6 @@
+import structlog
+
+
+class OriginLogger:
+    def __init__(self, origin: str):
+        self.logger = structlog.get_logger().bind(origin=origin)
```
aponcedeleonch, line 6: I think this needs to be
```
self.logger = structlog.get_logger("codegate").bind(origin=origin)
```

Or at least with this change I'm getting `debug` logs even though I set in the cli the logging level to INFO
```
2024-12-18 12:33:09 [debug    ] Processing body through pipeline: 0 bytes origin=copilot_proxy
2024-12-18 12:33:09 [debug    ] No pipeline strategy selected  origin=copilot_proxy
2024-12-18 12:33:09 [debug    ] Extracting path from /extensions/chat.json origin=copilot_proxy
2024-12-18 12:33:09 [debug    ] Processing body through pipeline: 0 bytes origin=copilot_proxy
2024-12-18 12:33:09 [debug    ] No pipeline strategy selected  origin=copilot_proxy
```

src/codegate/providers/copilot/provider.py
```diff
@@ -22,7 +23,8 @@
 )
 from codegate.providers.copilot.streaming import SSEProcessor
 
-logger = structlog.get_logger("codegate")
+setup_logging()
```
aponcedeleonch, line 26: Do we need this call to `setup_logging()`, I think it might reconfigure the logger  on-the-fly?

aponcedeleonch (approved):
Nothing, just tested works nicely!

src/codegate/providers/copilot/provider.py
yrobla: if we do not set it, copilot loses the scope and is not logging with the expected format

Issue #405, opened 2024-12-18
FIM not caching correctly JS files


aponcedeleonch:
The cache is working correctly for Python and languages which uses `#` as a character to start a comment. Languages that use `//` are not being registered in cache correctly. This is because the regex that we are using.

FIM Python request:
```python
# Some context
# my_path/my_file.py <-- Regex will match here

def foo():
   pass
```

FIM JS request
```javascript
// Some context
// my_path/my_file.js <-- Regex will not match here

...
```

Issue #365, opened 2024-12-16
Store metadata when alerting on secrets.


aponcedeleonch:
Currently, when secrets are recorded in the database, we are not capturing enough contextual information about them. Enhancing the recorded data can improve usability and visibility, particularly for display on the dashboard.

Record Contextual Information:
- Include the file name where the secret was detected.
- Capture the surrounding lines of code or text where the secret appears.
- This additional context will help users quickly identify and understand where the secret is coming from.

Classify Secrets by Type:
- Record the type of secret based on the matching regex (e.g., AWS secret, GitHub token, etc.).
- Since we already know which regex triggered the match, we can use this information to categorize the secret.

PR #408, opened 2024-12-18
fix: FIM not caching correctly non-python files

aponcedeleonch:
Closes: #405

The implemented cache was not working correctly because of the way the context is added for FIM requests in other programming languages which are not python. The context for the LLM is provided as single line comments. In Python this means lines which start with the character `#`. Other languages may have other starting sequence for single line comments, e.g. Javascript uses `//`. This PR changes the regex to detect the paths for other languages.


src/codegate/db/fim_cache.py
```diff
@@ -0,0 +1,136 @@
...
+
+        if self._is_cached_entry_old(context, cached_entry):
+            self._add_cache_entry(hash_key, context)
+            return True
+
+        if self._are_new_alerts_present(context, cached_entry):
+            self._add_cache_entry(hash_key, context)
+            return True
+
+        logger.info(f"FIM entry already in cache: {hash_key}.")
```
jhrozek, line 135: this should probably be a debug?

src/codegate/db/fim_cache.py
```diff
@@ -0,0 +1,136 @@
...
+        ]
+        new_cache = CachedFim(
+            timestamp=context.input_request.timestamp, critical_alerts=critical_alerts
+        )
+        self.cache[hash_key] = new_cache
+        logger.info(f"Added cache entry for hash key: {hash_key}")
+
+    def _are_new_alerts_present(self, context: PipelineContext, cached_entry: CachedFim) -> bool:
+        """Check if there are new alerts present"""
+        new_critical_alerts = [
```
jhrozek, line 102: I guess this would not alert if there was the same amount of alerts, but different? (This is probably an edge case..)

src/codegate/db/fim_cache.py
```diff
@@ -0,0 +1,136 @@
...
+        filepath = self._match_filepath(message, provider)
+        if filepath is None:
+            logger.warning("No path found in messages. Creating hash key from message.")
+            message_to_hash = f"{message}-{provider}"
+        else:
+            message_to_hash = f"{filepath}-{provider}"
+
+        logger.debug(f"Message to hash: {message_to_hash}")
+        hashed_content = hashlib.sha256(message_to_hash.encode("utf-8")).hexdigest()
+        logger.debug(f"Hashed contnet: {hashed_content}")
```
jhrozek, line 84: typo

src/codegate/db/fim_cache.py
```diff
@@ -0,0 +1,136 @@
...
+        # If no path is found, hash the entire prompt message.
+        if not matches:
+            return None
+
+        # Extract only the paths (2nd group from the match)
+        paths = [match[1] for match in matches]
+
+        # Copilot puts the path at the top of the file. Continue providers contain
+        # several paths, the one in which the fim is triggered is the last one.
+        if provider == "copilot":
```
jhrozek, line 68: I think it would be nicer if we had a provider interface instead of having a list of providers in the DB layer, but that's for later.

jhrozek (approved):
some nits here and there but looks good

src/codegate/db/fim_cache.py
aponcedeleonch: Agree! Actually, in the future we could think about transforming the DB stuff to a pipeline step.

src/codegate/db/fim_cache.py
aponcedeleonch: Yes, I was using it to debug while development. Will change

src/codegate/db/fim_cache.py
aponcedeleonch: Will fix

src/codegate/db/fim_cache.py
aponcedeleonch: Yes, I'm worried about the same thing. However, I couldn't come up with a better solution to distinguish between alerts across multiple pipeline runs for the moment, specially for differentiating secrets related alerts. Potentially, when we do #365, we could have a better solution.

jhrozek (approved):

Issue #402, opened 2024-12-17
Integration tests


jhrozek:
We have a decent unit test coverage, but we should build some integration tests as well, e.g. try talking to codegate as it passes requests to an LLM and then read the replies.

jhrozek:
Co-assigned @rdimitrov because he'll be looking at running the tests in CI

PR #411, opened 2024-12-18
Add initial set of integration tests

ptelang:
Related to #402

### DIFF
--- a/tests/integration/integration_tests.py
+++ b/tests/integration/integration_tests.py
@@ -1,0 +1,154 @@
+import asyncio
+import json
+import os
+import re
+
+import requests
+import structlog
+import yaml
+from dotenv import find_dotenv, load_dotenv
+from sklearn.metrics.pairwise import cosine_similarity
+
+from codegate.inference.inference_engine import LlamaCppInferenceEngine
+
+logger = structlog.get_logger("codegate")
+
+
+class CodegateTestRunner:
+    def __init__(self):
+        self.inference_engine = LlamaCppInferenceEngine()
+        self.embedding_model = "codegate_volume/models/all-minilm-L6-v2-q5_k_m.gguf"
+
+    @staticmethod
+    def call_codegate(url, headers, data):
+        response = None
+        try:
+            response = requests.post(url, headers=headers, json=data)
+        except Exception as e:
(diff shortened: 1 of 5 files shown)


tests/integration/integration_tests.py
```diff
@@ -0,0 +1,155 @@
+import asyncio
+import json
+import os
+import re
+
+import numpy as np
```
jhrozek, line 6: this import seems to be unused

tests/integration/integration_tests.py
```diff
@@ -0,0 +1,155 @@
...
+import requests
+import structlog
+import yaml
+from dotenv import find_dotenv, load_dotenv
+from sklearn.metrics.pairwise import cosine_similarity
```
jhrozek, line 11: should we add this import to `tool.poetry.group.dev.dependencies` in `pyproject.toml` ?

tests/integration/integration_tests.py
```diff
@@ -0,0 +1,155 @@
...
+
+from codegate.inference.inference_engine import LlamaCppInferenceEngine
+
+logger = structlog.get_logger("codegate")
+
+
+class CodegateTestRunner:
+    def __init__(self):
+        self.inference_engine = LlamaCppInferenceEngine()
+        self.embedding_model = "./models/all-minilm-L6-v2-q5_k_m.gguf"
```
jhrozek, line 21: since we already have the model at `codegate_volume/models/all-minilm-L6-v2-q5_k_m.gguf` in the repo why not point at that instead?

jhrozek (commented):
I think this is great and we can base the copilot tests on this, too!
I just have some questions inline about the imports and the models path, but those are details.

tests/integration/integration_tests.py
ptelang: Yes, it was a leftover. I removed it.

tests/integration/integration_tests.py
ptelang: Done.

tests/integration/integration_tests.py
ptelang: Yes, fixed.

jhrozek (approved):

Issue #401, opened 2024-12-17
Support other Copilot LLMs than OpenAI


jhrozek:
Our Copilot provider is fairly tied to it using OpenAI as LLM, but since recently, Copilot can use other LLMs as well. We should support that.

jhrozek:
Anthropic mostly works with copilot, except some issues around retrieving malicious packages in natural language chat.
The current hypothesis is that with anthropic in copilot, the system messages actually use `role: user` for whatever reason.

jhrozek:
I'll do this together with #415 because the free copilot also supports anthropic.

jhrozek:
PR #428 helps to extract package names and ecosystems with Copilot and Anthropic. There will be a follow up to match them in weaviate - we are discussing how to approach that with Pankaj. Once those are fixed, Anthropic will be supported with Copilot!

jhrozek:
I co-assigned @ptelang because during testing we saw issues with extracting package names from natural language prompts also with GPT and this needs an LLM expert :-)

jhrozek:
@blkt 👀

jhrozek:
I did re-test the Claude integration in the free copilot plan and everything works as expected now. Closing.

PR #428, opened 2024-12-19
Tune the system prompts to extract package names or ecosystems

jhrozek:
With Copilot+Anthropic the user message can contain instructions that
are passed from the chat client and really sound like a system message.
For whatever reason, the chat client includes them with `role=user`.
These instructions confuse the LLM when extracting the package names or
ecosystem and it replies with a mix of JSON and natural language.

To work around that, Pankaj suggested to tune the system prompt and tell
the LLM to ignore the instructions. This seems to work and helps
Anthropic extract the package names even with copilot user message.

Co-authored-by: Pankaj Telang <pankaj@stacklok.com>

Related: #401


ptelang (approved):
Looks good!

Issue #371, opened 2024-12-16
VSCode needs a restart to re-auth the user account via the proxy


lukehinds:
The session management in vscode is not very smart, we need to figure out a way around this.

PR #432, opened 2024-12-20
improve TLS handling with SNI support and cert caching

lukehinds:
Implements improved TLS handling system with Server Name Indication (SNI) support and certificate caching. Changes include:

    - Implement SNI support to provide better hostname verification
    - Add certificate expiration checks and automatic cleanup
    - Domain cert caching, with validation
    - Improved and refactored code layout

The new implementation improves performance by caching certificates and contexts
while providing proper hostname verification through SNI. This allows the proxy
to securely handle multiple GitHub domains with domain-specific certificates which in turn fixes: #371


src/codegate/ca/codegate_ca.py
```diff
@@ -1,18 +1,85 @@
...
+class TLSCertDomainManager:
+    """
+    This class manages SSL contexts for domain certificates with SNI
+    """
+
+    def __init__(self, ca_provider: "CertificateAuthority"):
+        self._ca = ca_provider
+        # Use strong references for caching
+        self._cert_cache: Dict[str, CachedCertificate] = {}
+        self._context_cache: Dict[str, ssl.SSLContext] = {}
```
jhrozek, line 41: This is not for this PR, but would we ever want to do some cache cleanup like least-recently used or are the certificates small enough that we don't care?

src/codegate/ca/codegate_ca.py
```diff
@@ -41,194 +108,179 @@ def __init__(self):
...
+            ca_cert_path = self.get_cert_path(Config.get_config().ca_cert)
+            logger.debug(f"Loading CA certificate for verification: {ca_cert_path}")
+            with open(ca_cert_path, "rb") as f:
+                ca_cert = x509.load_pem_x509_certificate(f.read(), default_backend())
+        except Exception as e:
+            logger.error(f"Failed to load CA certificate for verification: {e}")
+            return
+
+        # Get current time for expiry checks
+        current_time = datetime.now(timezone.utc)
```
jhrozek, line 134: I had to look up in the docs if datetime.now is equivalent to datetime.utcnow and based on my understanding it is - just wanted to double check during review if you think it's OK to use both in the code (not a blocker)

src/codegate/ca/codegate_ca.py
```diff
@@ -41,194 +108,179 @@ def __init__(self):
...
+                raise
+
+    def remove_certificates(self) -> None:
+        """Remove all cached certificates and contexts"""
+        logger.debug("Removing all cached certificates and contexts")
+        self.certs_dir = Config.get_config().certs_dir
+        # remove and recreate certs directory
+        try:
+            logger.debug(f"Removing certs directory: {self.certs_dir}")
+            os.rmdir(self.certs_dir)
```
jhrozek, line 235: This will only work if the certs_dir is empty, right? Or did you mean to use rmtree here?

jhrozek (approved):
This seems to work great and is an improvement. I left some small comments inline - two are questions and one might be a bug (the confusion between rmdir and rmtree).

But since this is quite a nice improvement, I say let's merge and if there are things to follow up on, let's make a follow-up patch.

src/codegate/ca/codegate_ca.py
lukehinds: I think so, I will follow up, just want to think it over a bit more

Issue #437, opened 2024-12-21
Review Weaviate as a vectorDB


lukehinds:
Review if weaviate is the right approach, we have seen some performance aspects impacting UX.

A possible alternative could be sqlite-vec.

PR #438, opened 2024-12-21
sqlite-vec vectorization database

lukehinds:
Migrate to sqlite-vec

This PR migrates from weaviate to sqlite-vec.

I tried to keep the logic flow the same as before. Initial tests show prompt augmented correctly after matching a search.

image

## How to test

Load the packages into the database:

```bash
poetry run python scripts/import_packages.py --jsonl-dir=data   
```

Select a package from data and test in chat " is malicious-crates-dummy from crates safe to use?"

You should then see:

```bash
2024-12-24T17:13:26.3dZ [debug    ] Found matching packages in sqlite-vec database matched_packages=['malicious-crates-dummy (crates)'] module=codegate pathname=/Users/lhinds/repos/stacklok/codegate-repos/codegate/src/codegate/pipeline/codegate_context_retriever/codegate.py
```

 
Closes: #437


src/codegate/storage/storage_engine.py
```diff
@@ -152,66 +130,99 @@ async def search(
...
-
-                response = collection.query.near_vector(
-                    query_vector[0],
+                query_embedding = np.array(query_vector[0], dtype=np.float32)
+                query_embedding_bytes = query_embedding.tobytes()
+
+                query_sql = """
+                    WITH distances AS (
+                        SELECT name, type, status, description,
+                               vss_distance(embedding, ?) as distance
```
ptelang, line 181: We need to use cosine distance here. I could not find documentation on vss_distance function, and not sure which kind of distance it computes. I see a vec_distance_cosine function which we can use instead.

https://alexgarcia.xyz/sqlite-vec/api-reference.html#vec_distance_cosine

src/codegate/storage/storage_engine.py
```diff
@@ -152,66 +130,99 @@ async def search(
...
+                    SELECT name, type, status, description, distance
+                    FROM distances
+                    WHERE distance <= ?
+                    ORDER BY distance ASC
+                    LIMIT ?
+                """  # nosec
+                logger.debug(
+                    "Performing vector similarity search",
+                    query=query,
+                    distance_threshold=distance,
```
ptelang, line 193: This query will require a full table scan. It will run slower with larger number of packages in the table. 
Unlike weaviate and other vector dbs, sqlite-vec does not support indexing.
https://alexgarcia.xyz/blog/2024/sqlite-vec-stable-release/index.html#only-brute-force-search-for-now

We will need to watch this and see how well it works with all of the bad packages data.

src/codegate/storage/storage_engine.py
lukehinds: Thanks ! well captured , as per slack:  I started out with sqlite-vss and then got a warning it was deprecated, so flipped to the sqlite-vec, but left the old similarity distance call in place which weirdly still worked

ptelang (dismissed):

ptelang (approved):

ptelang (approved):
Looks good!

PR #444, opened 2024-12-23
Remove package/ecosystem extraction using llm

ptelang:


lukehinds (approved):
DId not realise this was also using treesitter and proposing a replacement.

What was the situation with rust, any problems or just something to get around to later

lukehinds (approved):

ptelang:
> What was the situation with rust, any problems or just something to get around to later

I did not get around to adding rust. We can add it in the next iteration.

PR #463, opened 2024-12-28
Cline Support

lukehinds:
This should be considered experimental until tested more widely by the community.

I have it working with Anthropic and Ollama so far.


src/codegate/pipeline/secrets/signatures.py
```diff
@@ -215,16 +215,26 @@ def _load_signatures(cls) -> None:
...
+        """Search for secrets in the provided string or list of strings."""
         if not text:
             return []
 
         if not cls._yaml_path:
             raise RuntimeError("SecretFinder not initialized.")
 
+        # Convert list to string if necessary (needed for Cline, which sends a list of strings)
+        if isinstance(text, list):
+            text = "\n".join(str(line) for line in text)
```
JAORMX, line 228: This is converted to string to be later split to an array? Seems redundant. Maybe you want to pass it as-is if it's already an array?

src/codegate/providers/ollama/provider.py
```diff
@@ -46,12 +46,42 @@ def _setup_routes(self):
...
+            """
+            Special route for /api/tags that responds outside of the pipeline
+            Tags are used to get the list of models
+            https://github.com/ollama/ollama/blob/main/docs/api.md#list-local-models
+            """
+            async with httpx.AsyncClient() as client:
+                response = await client.get(f"{self.base_url}/api/tags")
+                return response.json()
+
+        @self.router.post(f"/{self.provider_route_name}/api/show")
```
yrobla, line 60: this seems unrelated to cline support?

lukehinds: The route or use of ollama?

src/codegate/pipeline/secrets/signatures.py
lukehinds: totally agree, trying to recall why I put that there now. Anyhow, removed in last commit as tested without it and all seems well

src/codegate/providers/ollama/provider.py
yrobla: ah i get it now, you add it because cline is needing additional endpoints that were not exposed before. I hit the same with Aider

JAORMX (approved):

aponcedeleonch (approved):
The code changes look good. I did a quick run through with Continue and Copilot and also all seems good. However, I don't have Cline to test on it

Issue #398, opened 2024-12-17
Prompt markdown sometimes is invalid


peppescg:
### Describe the issue

The dashboard ui is receiving an invalid markdown from prompts endpoint

The backtick should go into a paragraph 

### Steps to Reproduce

- ask a review adding a context file

the invalid markdown should be visible like that:
<img width="948" alt="Image" src="https://github.com/user-attachments/assets/4ecc33a3-70e9-44f8-9a80-e985921bd232" />

### Operating System

None

### Provider

None

PR #481, opened 2025-01-02
feat: remove context and query from messages endpoint

yrobla:
When sending the initial message to the dashboard endpoint, strip the initial prefixes with context and query

Closes: #398


peppescg (approved):
🚀

peppescg (approved):

Issue #475, opened 2024-12-31
Package Extractor fineses


lukehinds:
### Describe the issue

The Package Extractor code is assuming ecosystem somehow and that a word is actually a package. Managed to replicate this by using the words "how to code?"  into copilot chat:

<img width="587" alt="Image" src="https://github.com/user-attachments/assets/97871880-a389-4fea-8988-0f152e973593" />

A user in discord experience the same issue where the word column was extracted from a codesnippet and derived as a package in Rust:

```
import pandas as pd

def main():

    # Get the list of column names
    columns = df.columns.tolist()
```

cc @ptelang 

### Steps to Reproduce

Just type 'how to code?' into chat

### Operating System

Other

### IDE and Version

any

### Extension and Version

any

### Provider

GitHub Copilot

### Model

any

Issue #462, opened 2024-12-28
False Positive names triggering vector search


lukehinds:
### Describe the issue

While working on Cline support I noted that they often use the word 'vscode' in their prompts , referring to the IDE. This matches the following package and gets raised in CodeGate

https://www.insight.stacklok.com/report/npm/vscode

<img width="702" alt="Image" src="https://github.com/user-attachments/assets/57d18335-3ac0-44f9-a0c9-165ee319f0aa" />

### Steps to Reproduce

Create a prompt containing the word vscode , or any package name on the malicious / deprecated list used out of context of a package

<img width="702" alt="Image" src="https://github.com/user-attachments/assets/c2355c78-4228-4adb-ae5b-c9bacd375374" />

### Operating System

MacOS (Arm)

### IDE and Version

Version: 1.96.2 (Universal) Commit: fabdb6a30b49f79a7aba0f2ad9df9b399473380f

### Extension and Version

v3.0.8

### Provider

Ollama

### Model

v3.0.8

lukehinds:
@lukehinds / @ptelang to re-test to see if still valid

ptelang:
I tested and got the following:

<img width="835" alt="Image" src="https://github.com/user-attachments/assets/14831887-6cf2-48f5-add4-efe5a80807dd" />

ptelang:
Note in this case, Codegate doesn't know the programming language. Codegate's semantic search finds npm/vscode as a malicious package and adds it to the prompt context. Now, LLM fails to distinguish between "editor" and "package" and responds back with codegate detected vscode as malicious package.

PR #494, opened 2025-01-06
Reduce false positive matches by reducing the similarity distance

ptelang:
Related to #475 #462


jhrozek (dismissed):
I'm going to ack this and lean on your testing but in the future we should have automated tests that should tell us if a set of queries gives the expected results.

PR #503, opened 2025-01-07
fix: remove Authorization header from img publish

aponcedeleonch:
Since the repos are public we don't need the `Authorization` header anymore.


JAORMX (approved):

peppescg (approved):

lukehinds (changes requested):
block

lukehinds (approved):

PR #505, opened 2025-01-07
fix: invert the condition for certs output

yrobla:


src/codegate/cli.py
```diff
@@ -313,7 +313,7 @@ def serve(
         ca = CertificateAuthority.get_instance()
 
         certs_check = ca.check_and_ensure_certificates()
-        if certs_check:
+        if not certs_check:
```
lukehinds, line 316: Well spotted , may as well rename to `certs_exist`

lukehinds (approved):

PR #509, opened 2025-01-07
Fix copilot errors which cause client to hangup during FIM

ptelang:


src/codegate/providers/copilot/provider.py
```diff
@@ -764,6 +767,10 @@ def _proxy_transport_write(self, data: bytes):
             logger.error("Proxy transport not available")
             return
         self.proxy.transport.write(data)
+        # print("DEBUG =================================")
+        # print(data)
+        # print("DEBUG =================================")
```
ptelang, line 772: I kept these and another set of print statements for future debugging.

lukehinds:
Interesting, so I figure the lack of stream being True meant the socket would terminate early?

lukehinds (approved):
much better. I still see some connection drops, but this moves things forward

yrobla (approved):

Issue #508, opened 2025-01-07
Publish container images with release semver tag


danbarr:
We're publishing releases with semantic versions now, but the container images produced by each release don't have the version tag added. It would be nice to have these so users can optionally pin to a version, or easily compare behavior between versions.

I think this should just be a minor addition to the `docker/metadata-action` step.

PR #513, opened 2025-01-08
Add sem version as image tags

lukehinds:
This will result in the following tags:

`:latest`
`:v0`
`:v0.0`
`:v0.0.5`

Along with the existing ref tag

Closes: #508

### DIFF
--- a/.github/workflows/image-publish.yml
+++ b/.github/workflows/image-publish.yml
@@ -52,6 +52,9 @@ jobs:
             latest=true
           tags: |
             type=raw,value=${{ steps.version-string.outputs.tag }}
+            type=semver,pattern=v{{version}}
+            type=semver,pattern=v{{major}}
+            type=semver,pattern=v{{major}}.{{minor}}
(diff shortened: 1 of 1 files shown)


.github/workflows/image-publish.yml
```diff
@@ -52,6 +52,9 @@ jobs:
             latest=true
           tags: |
             type=raw,value=${{ steps.version-string.outputs.tag }}
+            type=semver,pattern=v{{version}}
+            type=semver,pattern=v{{major}}
+            type=semver,pattern=v{{major}}.{{minor}}
```
aponcedeleonch, line 57: aren't we also missing one line with `patch`
```
type=semver,pattern=v{{major}}.{{minor}}.{{patch}}
```

yrobla: we could discuss and add it later, depending at the level where we want to go

aponcedeleonch (approved):

Issue #512, opened 2025-01-08
Evaluate SQL approach - sqlc or raw approach


aponcedeleonch:
We're partially using sqlc right now. We're only using it to generate functions for `select` queries. Ideally we should use it to generate all the other queries as well: `insert` and `update`. However, at a first attempt we failed to make it work, specifically the `async` generated queries were not working out-of-the-box. We should evaluate if it's worth to keep sqlc or drop it completely in favor of having raw SQL queries.

aponcedeleonch:
No matter what I do I am getting the following error when trying to `insert` something using sqlc auto-generated code
```log
error=(sqlite3.ProgrammingError) Incorrect number of bindings supplied. The current statement uses 5, and there are 0 supplied.
[SQL: -- name: create_prompt :one
INSERT INTO prompts (
          id, timestamp, provider, request, type
) VALUES (
  ?, ?, ?, ?, ?
)
RETURNING id, timestamp, provider, request, type
]
(Background on this error at: https://sqlalche.me/e/20/f405)
```

I believe the error comes from the fact that the auto-generated code is trying to use a `dict` as input, note the `p*` keys using as values the attributes of the function.
```python
def create_prompt(self, *, id: Any, timestamp: Any, provider: Optional[Any], request: Any, type: Any) -> None:
        self._conn.execute(sqlalchemy.text(CREATE_PROMPT), {
            "p1": id,
            "p2": timestamp,
            "p3": provider,
            "p4": request,
            "p5": type,
        })
```

For the above to work, `CREATE_PROMPT` should be the following. Note the `p*` instead of `?` which is what the auto-generated code contains.
```python
CREATE_PROMPT = """-- name: create_prompt \\:one
INSERT INTO prompts (
          id, timestamp, provider, request, type
) VALUES (
  :p1, :p2, :p3, :p4, :p5
)
RETURNING id, timestamp, provider, request, type
""""
```

PR #515, opened 2025-01-08
Remove sqlc and its classes

aponcedeleonch:
Closes: #512

This PR drops sqlc related-files and adds the needed code to use directly SQLAlchemy. It most parts of the code we were already using it, hence the changes are not so drastic. See #512 issue for the reasoining on why to remove `sqlc`.


lukehinds (approved):
looks good. we might want to look at introducing connection pools , but not important as its one-to-one, but might help when their are a lot of prompts / outputs occuring with a high frequency.

PR #516, opened 2025-01-08
fix: add monitoring for idle connections and close them

yrobla:
need to find a better way


lukehinds (commented):
Could add a bit more context to what the change introduces, I figure these are asyncio methods, is 10 seconds sufficient? Perhaps there might be cases where the model is calculating a large prompt and exceeds that value?

lukehinds (approved):

Issue #517, opened 2025-01-08
copilot provider doesn't handle body arriving in a different message than the headers


jhrozek:
in _forward_data_to_target in the copilot pipeline we pass the data through the pipeline always even if we only receive headers in one message and the body in the subsequent message. This causes the first request (headers) to bypass the pipeline because the body is empty and the request to fail in case e.g. the server checks if the body is valid JSON. 

Since our pipeline needs the complete body of the request anyway, we need to buffer the whole body before forwarding it to the pipeline (if we didn't have the pipeline, we could probably just start writing the body in chunks as soon as the first piece of the body arrived)

PR #523, opened 2025-01-09
Add first two copilot integration tests

jhrozek:
- **Add invokehttp to dummy data** - We'd like to write unit tests that would ideally work equally well with the released docker container as well as for local development. Let's add a well-known malicious library to the data set for testing.
- **Add data types for pipeline routing, add route for OpenAI** - We just had a simple if-else for pipeline routing. Instead, let's add a data type that includes the paths and optionally a `target_url`. We can use that to add a routing for OpenAI which we'll use for testing the Copilot provider. In the future we can make the routings pluggable.
- **When processing forwarded HTTP request, don't process until we have received the whole body** - It might happen that the proxied request arrives in two chunks - first just the headers and then the body. In that case we would have sent just the headers with an empty body through the pipeline which might trigger errors like "400 Client Error", then the body would arrive in the next chunk. This patch changes the logic to keep the whole body in the buffer until it is complete and can be processed by the pipeline and sent off to the server.

- **Add integration tests for the copilot provider** - 

Since the copilot provider is a proxy, we add a "requester" module that
depending on the provider makes a request either using raw python
requests like earlier or by setting a proxy and using a CA cert file.

To be able to add more tests, we also add more kinds of checks, in
addition to the existing one which makes sure the reply is like the
expected one using cosine distance, we also add checks that make sure
the LLM reply contains or doesn't contain a string.

We use those to add a test that ensures that the copilot provider chat
works and that the copilot chat refuses to generate code snippet with a
malicious package.

To be able to run a subset of tests, we also add the ability to select a
subset of tests based on a provider (`codegate_providers`) or the test name (`codegate_test_names`)

These serve as the base for further integration tests.

To run them, call:
```
CODEGATE_PROVIDERS=copilot \
CA_CERT_FILE=/Users/you/devel/codegate/codegate_volume/certs/ca.crt \
ENV_COPILOT_KEY=your-openapi-key \
python tests/integration/integration_tests.py
```

Related: #402
Fixes: #517

### DIFF
--- a/data/malicious.jsonl
+++ b/data/malicious.jsonl
@@ -8,3 +8,4 @@
 {"name":"malicious-go-dummy","type":"go","description":"Dummy malicious to test with simple package name on go"}
 {"name":"@prefix/malicious-crates-dummy","type":"crates","description":"Dummy malicious to test with encoded package name on crates"}
 {"name":"malicious-crates-dummy","type":"crates","description":"Dummy malicious to test with simple package name on crates"}
+{"name":"invokehttp","type":"pypi","description":"Invokehttp is a malicious package"}
--- a/src/codegate/providers/copilot/mapping.py
+++ b/src/codegate/providers/copilot/mapping.py
@@ -1,4 +1,6 @@
-from typing import List
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Optional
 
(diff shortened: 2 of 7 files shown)


ptelang:
Looks great! Thanks for refactoring the test framework code.

ptelang (approved):

Issue #528, opened 2025-01-09
Secrets redaction response not working for Copilot in v0.1.5


danbarr:
### Describe the issue

Starting in v0.1.5, the secrets redaction message "CodeGate prevented X secrets from being leaked" message is not inserted into explain/chat responses anymore, and encrypted secrets in the response are not decrypted as expected. They are still detected/encrypted on the way out.

This is only affecting Copilot. Using Continue, the message is added as expected.

Revered to v0.1.4 and it's working as expected, and I can reproduce reliably on v0.1.5.

v0.1.5:
<img width="770" alt="Image" src="https://github.com/user-attachments/assets/c492716f-77f5-4ceb-9926-9f6c263bb93e" />

v0.1.4, expected response:
<img width="770" alt="Image" src="https://github.com/user-attachments/assets/02e849ce-fd0f-430b-ab3a-5386bf19b577" />

### Steps to Reproduce

With CodeGate v0.1.5, use Copilot Explain or Chat with the conf.ini file from the codegate-demonstration repo.

### Operating System

MacOS (Arm)

### IDE and Version

VS Code 1.96.2

### Extension and Version

GitHub Copilot 1.256.0

### Provider

GitHub Copilot

### Model

GPT-4

PR #547, opened 2025-01-10
Fix pipeline handling in the copilot provider

jhrozek:
- **Create the pipeline sensitive context when creating a pipeline instance, not on every processing**
- **Create pipeline instance when creating the SequentialPipelineProcessor not for every process**
- **Create the pipelines only once in the copilot provider**
- **test me: is this needed**

Fixes: #528

### DIFF
--- a/src/codegate/pipeline/base.py
+++ b/src/codegate/pipeline/base.py
@@ -340,8 +340,9 @@ class SequentialPipelineProcessor:
         self.pipeline_steps = pipeline_steps
         self.secret_manager = secret_manager
         self.is_fim = is_fim
+        self.instance = self._create_instance()
 
-    def create_instance(self) -> InputPipelineInstance:
+    def _create_instance(self) -> InputPipelineInstance:
         """Create a new pipeline instance for processing a request"""
         return InputPipelineInstance(self.pipeline_steps, self.secret_manager, self.is_fim)
 
@@ -356,7 +357,6 @@ class SequentialPipelineProcessor:
         is_copilot: bool = False,
     ) -> PipelineResult:
         """Create a new pipeline instance and process the request"""
-        instance = self.create_instance()
-        return await instance.process_request(
+        return await self.instance.process_request(
             request, provider, model, api_key, api_base, extra_headers, is_copilot
         )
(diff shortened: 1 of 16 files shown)


src/codegate/pipeline/base.py
```diff
@@ -275,6 +275,13 @@ def __init__(
         self.secret_manager = secret_manager
         self.is_fim = is_fim
         self.context = PipelineContext()
+
+        # we create the sesitive context here so that it is not shared between individual requests
+        # TODO: could we get away with just generating the session ID for an instance?
```
ptelang, line 280: Assuming each request has exactly one instance of the pipeline and each pipeline has one instance of PipelineSensitiveData, the session ID can be generated for the pipeline instance and used in PipelineSensitiveData.

src/codegate/providers/copilot/provider.py
```diff
@@ -150,8 +150,16 @@ def __init__(self, loop: asyncio.AbstractEventLoop):
         self.cert_manager = TLSCertDomainManager(self.ca)
         self._closing = False
         self.pipeline_factory = PipelineFactory(SecretsManager())
+        self.input_pipeline: Optional[CopilotPipeline] = None
+        self.fim_pipeline: Optional[CopilotPipeline] = None
+        # the context as provided by the pipeline
         self.context_tracking: Optional[PipelineContext] = None
 
+    def _ensure_pipelines(self):
+        if not self.input_pipeline or not self.fim_pipeline:
```
ptelang, line 159: Do we need to instantiate both the pipelines here, or can we just create one conditionally based on if the request is chat or fim?

src/codegate/providers/copilot/provider.py
```diff
@@ -288,6 +295,8 @@ async def _forward_data_through_pipeline(self, data: bytes) -> Union[HttpRequest
             http_request.headers,
             http_request.body,
         )
+        # TODO: it's weird that we're overwriting the context. Should we set the context once? Maybe when
+        # creating the pipeline instance?
```
ptelang, line 299: yes, that makes sense.

jhrozek:
No longer a draft and can be tested/reviewed, but only now did I notice @ptelang 's comments (sorry I was focused on getting the damn thing working first).
I'll fix them first thing in the morning.

src/codegate/providers/copilot/provider.py
jhrozek: we could, I'll add that to the issue I'll open

lukehinds (approved):

lukehinds (approved):

Issue #519, opened 2025-01-08
Secrets are double-counted via Copilot Explain function


danbarr:
### Describe the issue

Secrets detected in code are being double-counted when I use the Copilot Explain feature (select some code, right-click and Copilot->Explain). This does not happen if I send the same file as context in a Chat prompt.

Video shows the expected result via Chat (2 secrets detected), then the double-result via Explain (4 secrets detected):

https://github.com/user-attachments/assets/9c697b97-072b-402d-a350-59db6d176e5d

### Steps to Reproduce

See video - I used the conf.ini file in [codegate-demonstration](https://github.com/stacklok/codegate-demonstration). 

### Operating System

MacOS (Arm)

### IDE and Version

VS Code 1.96.2

### Extension and Version

GitHub Copilot 1.256.0

### Provider

GitHub Copilot

### Model

GPT-4

### Logs

````
2025-01-08T22:29:38.008Z [debug    ] Processing body through pipeline: 5483 bytes module=provider origin=copilot_proxy pathname=/app/src/codegate/providers/copilot/provider.py
2025-01-08T22:29:38.008Z [debug    ] Found 2 secrets in the user message module=secrets pathname=/app/src/codegate/pipeline/secrets/secrets.py
2025-01-08T22:29:38.008Z [info     ]
Found secrets:                module=secrets pathname=/app/src/codegate/pipeline/secrets/secrets.py
2025-01-08T22:29:38.008Z [debug    ] Stored secret                  encrypted=qsuXopnOcxUETo0zhMclZoYijjGri86kTvIqAjZhfcfk5VCTeI/xsI+EObT3WJY8aJx3uAx6afFuBs4= module=manager pathname=/app/src/codegate/pipeline/secrets/manager.py service=Amazon type=Access Key
2025-01-08T22:29:38.008Z [info     ]
Service: Amazon
Type: Access Key
Original: AKIAIOSFODNN7EXAMPLE
Encrypted: REDACTED<$qsuXopnOcxUETo0zhMclZoYijjGri86kTvIqAjZhfcfk5VCTeI/xsI+EObT3WJY8aJx3uAx6afFuBs4=> module=secrets pathname=/app/src/codegate/pipeline/secrets/secrets.py
[...]

PR #581, opened 2025-01-14
fix: Do not double count secrets on `/explain` copilot function

aponcedeleonch:
Closes: #519

When using `/explain` function in copilot we were double counting th secrets. The problem was that we were getting several `user` messages after the last `assistant` message. We are using the last `assistant` message as means to identify the user messages. Here is an example of how the request looked like for `/explain`:
```
[
  ...,
  {"role": "assistant", "content": "some content"},
  {"role": "user", "content": content_with_secrets},
  {"role": "user", "content": content_with_secrets},
]
```

To avoid double counting now we check which was the secrets that matched after the last `assistant` message and only consider the unique ones.

### DIFF
--- a/src/codegate/pipeline/secrets/secrets.py
+++ b/src/codegate/pipeline/secrets/secrets.py
@@ -106,10 +106,10 @@ class SecretsModifier:
         end_line = min(secret_line + surrounding_lines, len(lines))
         return "\n".join(lines[start_line:end_line])
 
-    def obfuscate(self, text: str) -> tuple[str, int]:
+    def obfuscate(self, text: str) -> tuple[str, List[Match]]:
         matches = CodegateSignatures.find_in_string(text)
         if not matches:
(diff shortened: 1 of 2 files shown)


src/codegate/pipeline/secrets/secrets.py
```diff
@@ -371,8 +373,6 @@ async def process_chunk(
             if original_value is None:
                 # If value not found, leave as is
                 original_value = match.group(0)  # Keep the REDACTED marker
-            else:
```
jhrozek, line 374: oops sorry this is some leftover debugging by me

jhrozek (approved):

src/codegate/pipeline/secrets/secrets.py
aponcedeleonch: Yes, I assumed so. No problem :)

Issue #454, opened 2024-12-26
Codegate Projects (repos)


lukehinds:
projects is something that myself and @jhrozek  discussed, and the more our discussion unfolded, the more it was realised how useful this could be in unlocking many features that would be highly beneficial to users. It is also an ‘inevitable’ that we will need such a feature, so sooner the better and less painful in the long run. 

This document attempts to describe the potential scope of Projects, what might they look like, and what they could unlock feature-wise.

## Non-Goals

Let’s start off with non-goals, as this is important. Projects would not mean a multi-tenant system with access control, hierarchy etc. If we need that (which will be much later) then a system such as Minder would be a more logical choice. 

## So what is a Project?

A project (in respect of this document) is essentially a means of grouping elements under a single codebase / repository. 

At present, we track the following elements, which are collated together as singular object:

- Prompt (chat or FIM based, either from the user or the IDE / Agent)  
- LLM Output  
- Secrets  
- Malicious / Deprecated Packages

These can be further broken down into pipelines, where Secrets and Malicious Packages are inserted into a flow of checks via the pipeline system.

Over time, we plan to introduce more elements, such as token count, suspicious shell patterns, and folder obfuscation etc. It’s at this point that you can start to see a large amount of data all sitting in the same pile, making for an organized mess in the dashboard.

A project is a representation of a git repository, perhaps mounted in a volume or by other means (TBD)

The first obvious benefit of Projects in CodeGate is the grouping together of elements, neatly under a parent (a repository). This will make the UX far more user-friendly than a big long ever increasing list of prompts, secrets and malicious packages.

## Root Element (Repository)

The Nexus of a Project is the **Repository**. This is a logical component to use, for the following reasons:

1. Prompts are specific to a codebase / repository (outside of generic chatbot scenarios)  
2. Outputs are specific to a codebase / repository   
3. Files and their paths are specific to a codebase / repository  
4. Secrets and Malicious packages are are specific to a codebase / repository

So we see we have logical grouping with a repository. 

With a repository as the root element / grouping and local file access of the repository being made available to codegate, this then allows some highly useful features

### Better UI Organization

Right now we lump everything in together, with separate buckets for each repo, it will be far more useful to helping users find what they need. 

### Project Based System Prompt

[...]

PR #583, opened 2025-01-14
WIP: feat: Initial code to load workspaces from a specific container path

aponcedeleonch:
Related: #454

This is the initial work to create workspaces when the server is initialized. The idea is that the user mounts a volume at the specific location: `/app/codegate_workspaces` and read from there the git repositories.

### DIFF
--- a/src/codegate/config.py
+++ b/src/codegate/config.py
@@ -54,6 +54,9 @@ class Config:
     force_certs: bool = False
 
     max_fim_hash_lifetime: int = 60 * 5  # Time in seconds. Default is 5 minutes.
+    ignore_paths_workspaces = [
+        ".git", "__pycache__", ".venv", ".DS_Store", "node_modules", ".pytest_cache", ".ruff_cache"
+    ]
 
     # Provider URLs with defaults
     provider_urls: Dict[str, str] = field(default_factory=lambda: DEFAULT_PROVIDER_URLS.copy())
--- a/src/codegate/workspaces/workspaces.py
+++ b/src/codegate/workspaces/workspaces.py
@@ -1,0 +1,80 @@
+import asyncio
+import json
+import uuid
+from pathlib import Path
+from typing import Dict, List, Optional, Union
+
+from pydantic import BaseModel
+
+from codegate.db.connection import DbRecorder
+from codegate.db.models import Workspace
+
+
+class Folder(BaseModel):
+    files: List[str] = []
+
+
+class Repository(BaseModel):
+    name: str
+    folder_tree: Dict[str, Folder]
+
+
+class FolderRepoScanner:
+
+    def __init__(self, ignore_paths: Optional[List[str]] = None):
+        if ignore_paths is None:
+            ignore_paths = []
+        self.ignore_paths = ignore_paths
+
+    def _should_skip(self, path: Path):
+        """Skip certain paths that are not relevant for scanning."""
+        return any(part in path.parts for part in self.ignore_paths)
+
+    def _read_repository_structure(self, repo_path: Path) -> Dict[str, Folder]:
+        folder_tree: Dict[str, Folder] = {}
+        for path in repo_path.rglob('*'):
+            if self._should_skip(path):
+                continue
+
+            relative_path = path.relative_to(repo_path)
+            if path.is_dir():
+                folder_tree[str(relative_path)] = Folder()
+            else:
+                parent_dir = str(relative_path.parent)
+                if parent_dir not in folder_tree:
+                    folder_tree[parent_dir] = Folder()
+                folder_tree[parent_dir].files.append(path.name)
+        return folder_tree
+
+    def read(self, path_str: Union[str, Path]) -> List[Repository]:
+        path_dir = Path(path_str)
+        if not path_dir.is_dir():
+            print(f"Path {path_dir} is not a directory")
+            return []
+
+        found_repos = []
+        for child_path in path_dir.rglob('*'):
+            if child_path.is_dir() and (child_path / ".git").exists():
+                repo_structure = self._read_repository_structure(child_path)
+                new_repo = Repository(name=child_path.name, folder_tree=repo_structure)
+                found_repos.append(new_repo)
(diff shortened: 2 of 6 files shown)


src/codegate/config.py
```diff
@@ -54,6 +54,9 @@ class Config:
     force_certs: bool = False
 
     max_fim_hash_lifetime: int = 60 * 5  # Time in seconds. Default is 5 minutes.
+    ignore_paths_workspaces = [
+        ".git", "__pycache__", ".venv", ".DS_Store", "node_modules", ".pytest_cache", ".ruff_cache"
+    ]
```
jhrozek, line 59: would it make sense to just include the contents of gitignore?
Sounds like we should make this configurable down the road.

src/codegate/workspaces/workspaces.py
```diff
@@ -0,0 +1,80 @@
...
+class Workspaces:
+
+    def __init__(self):
+        self._db_recorder = DbRecorder()
+
+    def read_workspaces(self, path: str, ignore_paths: Optional[List[str]] = None) -> None:
+        repos = FolderRepoScanner(ignore_paths).read(path)
+        workspaces = [
+            Workspace(
+                id=str(uuid.uuid4()),
```
jhrozek, line 74: Does this mean that if I restart the container I get a new Workspace per repo in the tree?

src/codegate/workspaces/workspaces.py
```diff
@@ -0,0 +1,80 @@
...
+from codegate.db.models import Workspace
+
+
+class Folder(BaseModel):
+    files: List[str] = []
+
+
+class Repository(BaseModel):
+    name: str
+    folder_tree: Dict[str, Folder]
```
JAORMX, line 19: Is the intent to store the whole directory tree of a repository?

src/codegate/config.py
aponcedeleonch: I did consider the contents of `.gitignore` but if we use that it would mean skipping files that may contain secrets but could still leaked to LLMs. 

I was planning to make `ignore_paths_workspaces` configurable through the cli, I just didn't have time to do so. The values here would be the defaults

src/codegate/workspaces/workspaces.py
aponcedeleonch: Yes, I still need to add functionality to avoid creating a new workspace if the repo already exists.

src/codegate/workspaces/workspaces.py
aponcedeleonch: Yes, the intent is to store the whole directory tree of a repository. The reasoning behind it is to do fast lookups when we see a path in the received code snippets. Right now, we get the path of a code snippet if it was supplied for context to the LLM. Example:
```json
{
  "messages": [
    {
      "role": "user",
      "content": "\n\n```py codegate/src/codegate/pipeline/factory.py (1-57)\nfrom typing import List\n\nfrom codegate.config import Config\nfrom codegate.pipeline.base import PipelineStep, SequentialPipelineProcessor\nfrom codegate.pipeline.codegate_context_retriever.codegate import CodegateContextRetriever\nfrom codegate.pipeline.extract_snippets.extract_snippets import CodeSnippetExtractor\nfrom codegate.pipeline.extract_snippets.output import CodeCommentStep\nfrom codegate.pipeline.output import OutputPipelineProcessor, OutputPipelineStep\nfrom codegate.pipeline.secrets.manager import SecretsManager\nfrom codegate.pipeline.secrets.secrets import (\n    CodegateSecrets,\n    SecretRedactionNotifier,\n    SecretUnredactionStep,\n)\nfrom codegate.pipeline.system_prompt.codegate import SystemPrompt\nfrom codegate.pipeline.version.version import CodegateVersion\n\n\nclass PipelineFactory:\n    def __init__(self, secrets_manager: SecretsManager):\n        self.secrets_manager = secrets_manager\n\n    def create_input_pipeline(self) -> SequentialPipelineProcessor:\n        input_steps: List[PipelineStep] = [\n            # make sure that this step is always first in the pipeline\n            # the other steps might send the request to a LLM for it to be analyzed\n            # and without obfuscating the secrets, we'd leak the secrets during those\n            # later steps\n            CodegateSecrets(),\n            CodegateVersion(),\n            CodeSnippetExtractor(),\n            CodegateContextRetriever(),\n            SystemPrompt(Config.get_config().prompts.default_chat),\n        ]\n        return SequentialPipelineProcessor(input_steps, self.secrets_manager, is_fim=False)\n\n    def create_fim_pipeline(self) -> SequentialPipelineProcessor:\n        fim_steps: List[PipelineStep] = [\n            CodegateSecrets(),\n        ]\n        return SequentialPipelineProcessor(fim_steps, self.secrets_manager, is_fim=True)\n\n    def create_output_pipeline(self) -> OutputPipelineProcessor:\n        output_steps: List[OutputPipelineStep] = [\n            SecretRedactionNotifier(),\n            SecretUnredactionStep(),\n            CodeCommentStep(),\n        ]\n        return OutputPipelineProcessor(output_steps)\n\n    def create_fim_output_pipeline(self) -> OutputPipelineProcessor:\n        fim_output_steps: List[OutputPipelineStep] = [\n            # temporarily disabled\n            # SecretUnredactionStep(),\n        ]\n        return OutputPipelineProcessor(fim_output_steps)\n\n```\nwhats this code doing?"
    }
  ],
  "model": "hosted_vllm/unsloth/Qwen2.5-Coder-32B-Instruct",
  "max_tokens": 4096,
  "stream": true,
  "base_url": "https://inference.codegate.ai/v1"
}
```

JAORMX: gotcha, that makes sense. Let's give this a little bit of thought; every time a file is added/removed we'd have to rewrite the JSON blob into the database and that's not optimal either.

aponcedeleonch:
The effort to automatically detect repositories from the information provided by the client is stopped. At the moment we don't have enough information to accurately pin-point the repository in which a user is working on. The workspaces effort will continue with #600

PR #589, opened 2025-01-15
Add all-in-one minimal nginx config

JAORMX:
The nginx configuration uses a lot of defaults that work well in
server-like environments. However, this is not optimized for a local
dev-environment or single-container as codegate is meant to run.

The nginx server spawns several processes which mostly remain idle. This
is wasteful and it is taking over resources that could be used for
something else. The intention of this new configuration is to minimize
footprint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -78,7 +78,7 @@ RUN chown -R codegate /var/lib/nginx && \
     chown -R codegate /var/log/nginx && \
     chown -R codegate /run
 
-COPY nginx.conf /etc/nginx/conf.d/default.conf
+COPY nginx.conf /etc/nginx/nginx.conf
 
 # Remove include /etc/nginx/sites-enabled/*; from the default nginx.conf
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -78,7 +78,7 @@ RUN chown -R codegate /var/lib/nginx && \
     chown -R codegate /var/log/nginx && \
     chown -R codegate /run
 
-COPY nginx.conf /etc/nginx/conf.d/default.conf
+COPY nginx.conf /etc/nginx/nginx.conf
```
jhrozek, line 81: is there any system-default nginx.conf? Which of them takes precedence? Just making sure we won't block ourselves from receiving changes to defaults in the future

jhrozek (approved):

Dockerfile
JAORMX: right, that's exactly what I'm overwriting here. We no longer take into account the default nginx.conf and instead write our own.

PR #593, opened 2025-01-15
feat: Introduce DB migrations

aponcedeleonch:
Related: #583

We had been using a single DB schema that didn't change until now. This introduces migrations using `alembic`. To create a new migration one can use:
```sh
alembic revision -m "My migration"
```
That should generate an empty migration file that needs to be hand-filled. Specifically the `upgrade` method which will be the one executed when running the migration. See the existent migration `30d0144e1a50_init_db.py` for a filled example.
```python
"""My migration

Revision ID: <some_hash>
Revises: <previous_hash>
Create Date: YYYY-MM-DD HH:MM:SS.XXXXXX
"""
from alembic import op
import sqlalchemy as sa

revision = '<some_hash>'
down_revision = '<previous_hash>'
branch_labels = None
depends_on = None

def upgrade():
    pass

def downgrade():
    pass
```


alembic.ini
```diff
@@ -0,0 +1,82 @@
+# A generic, single database configuration.
```
JAORMX, line 1: nit: Can we remove all the useless comments from this file?

migrations/versions/30d0144e1a50_init_db.py
```diff
@@ -0,0 +1,83 @@
...
+            other_settings TEXT  -- JSON stored as TEXT
+        );
+        """
+    )
+    # Create indexes for foreign keys and frequently queried columns
+    op.execute("CREATE INDEX idx_outputs_prompt_id ON outputs(prompt_id);")
+    op.execute("CREATE INDEX idx_alerts_prompt_id ON alerts(prompt_id);")
+    op.execute("CREATE INDEX idx_prompts_timestamp ON prompts(timestamp);")
+    op.execute("CREATE INDEX idx_outputs_timestamp ON outputs(timestamp);")
+    op.execute("CREATE INDEX idx_alerts_timestamp ON alerts(timestamp);")
```
JAORMX, line 79: For my education, why is everything executed in dedicated `op.execute` calls? Is this a best practice in these scripts? I'm acquainted with alembic migrations, but I'm more used to using it alongside an ORM.

JAORMX (dismissed):

alembic.ini
aponcedeleonch: Yes, I was 50-50 myself between keeping them or remove them. Now they're removed

JAORMX (approved):

PR #595, opened 2025-01-15
Log to inspect incoming requests

aponcedeleonch:
This is not meant to be merged but as temporary means for logging incoming requests

### DIFF
--- a/src/codegate/providers/base.py
+++ b/src/codegate/providers/base.py
@@ -212,6 +212,9 @@ class BaseProvider(ABC):
         - Execute the completion and translate the response back to the
           provider-specific format
         """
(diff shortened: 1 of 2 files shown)


src/codegate/providers/base.py
```diff
@@ -212,6 +212,9 @@ async def complete(
         - Execute the completion and translate the response back to the
           provider-specific format
         """
+        # Uncomment to check the incoming data
+        import json
+        logger.info(f"Received request:\n{json.dumps(data)}")
```
JAORMX, line 217: Should this be debug?

aponcedeleonch:
The debug that was needed from branch was completed.

Issue #586, opened 2025-01-14
Error with Aider and Ollama


danbarr:
### Describe the issue

Trying to use CodeGate with Aider + Ollama, I get the following error on every interaction:

```plain
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Aider v0.71.1
Model: ollama/qwen2.5-coder:7b with whole edit format
Git repo: .git with 19 files
Repo-map: using 4096.0 tokens, auto refresh
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
> codegate-version

litellm.APIConnectionError: Error parsing chunk: Expecting value: line 1 column 1 (char 0),
Received chunk:
Traceback (most recent call last):
  File "/opt/homebrew/Cellar/aider/0.71.1/libexec/lib/python3.12/site-packages/litellm/llms/base_llm/base_model_iterator.py", line 74, in __next__
    return self._handle_string_chunk(str_line=str_line)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/aider/0.71.1/libexec/lib/python3.12/site-packages/litellm/llms/ollama/completion/transformation.py", line 390, in
_handle_string_chunk
    return self.chunk_parser(json.loads(str_line))
                             ^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.8/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.8/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/decoder.py", line 338, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.12/3.12.8/Frameworks/Python.framework/Versions/3.12/lib/python3.12/json/decoder.py", line 356, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

[...]

PR #596, opened 2025-01-15
fix: correct problems with ollama and aider

yrobla:
It has a pair of problems:
- always need to check that the model contains some value
- only send chunks that contain data, and do not send double newlines as they get considered as a different chunk

Closes: #586

### DIFF
--- a/src/codegate/providers/ollama/completion_handler.py
+++ b/src/codegate/providers/ollama/completion_handler.py
@@ -1,3 +1,4 @@
+import json
 from typing import AsyncIterator, Optional, Union
 
 import structlog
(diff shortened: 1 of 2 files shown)


aponcedeleonch (approved):

danbarr:
This resolved the errors I saw with Ollama, however I get pretty unexpected results:

image

danbarr (dismissed):
I think this did something to malicious package detection. Running off this branch, malicious packages aren't being detected at all for me anymore.

jhrozek:
> I think this did something to malicious package detection. Running off this branch, malicious packages aren't being detected at all for me anymore.

Note that if you're running from source, the full malicious database is not available, that's only added when producing the release image. Only packages from https://github.com/stacklok/codegate/blob/main/data/malicious.jsonl and other files in that directory are.

yrobla:
rebased on top of latest patches, and testing again

PR #600, opened 2025-01-15
feat: Initial migration for Workspaces and pipeline step

aponcedeleonch:
Related: #454

We noticed most of the incoming-requests which contain a code-snippet only list a relative path with respect to where the code editor is opened. This would make difficult to accurately distinguish between repositories in Codegate. For example, a user could open 2 different code Python repositorites in different session and both repositories contain a `pyproject.toml`. It would be impossible for Codegate to determine the real repository of the file only using the relative path.

Hence, the initial implementation of Workspaces will rely on a pipeline step that is able to take commands a process them. Some commands could be:
- List workspaces
- Add workspace
- Switch active workspace
- Delete workspace

It would be up to the user to select the desired active workspace.

This PR introduces an initial migration for Workspaces and the pipeline step with the `list` command.


migrations/versions/5c2f3eee5f90_introduce_workspaces.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def upgrade() -> None:
+    # Workspaces table
+    op.execute(
+        """
+        CREATE TABLE workspaces (
+            id TEXT PRIMARY KEY,  -- UUID stored as TEXT
+            name TEXT NOT NULL,
+            is_active BOOLEAN NOT NULL DEFAULT 0
```
JAORMX, line 27: In data corruption cases or a race condition, we'd end up with an undefined state for which workspace is active with this schema.

Can we create a "sessions" table instead? The session would have a UUID as identifier, and in an initial implementation it would be a singleton. We can then start a session as soon as codegate gets a request. The session would hold the workspace identifier. Wdyt?

migrations/versions/5c2f3eee5f90_introduce_workspaces.py
```diff
@@ -0,0 +1,40 @@
...
+        );
+        """
+    )
+    op.execute("INSERT INTO workspaces (id, name, is_active) VALUES ('1', 'default', 1);")
+    # Alter table prompts
+    op.execute("ALTER TABLE prompts ADD COLUMN workspace_id TEXT REFERENCES workspaces(id);")
+    op.execute("UPDATE prompts SET workspace_id = '1';")
+    # Create index for workspace_id
+    op.execute("CREATE INDEX idx_prompts_workspace_id ON prompts (workspace_id);")
```
JAORMX, line 36: We'll be accessing workspaces by name. Should we create an unique index for names as well? I also cannot see a use case for workspaces with the same name.

src/codegate/pipeline/workspace/workspace.py
```diff
@@ -0,0 +1,106 @@
...
+
+    async def _list_workspaces(self, *args):
+        """
+        List all workspaces
+        """
+        workspaces = await self._db_recorder.get_workspaces()
+        print(workspaces)
+        respond_str = ""
+        for workspace in workspaces:
+            respond_str += f"{workspace.id} - {workspace.name}"
```
JAORMX, line 29: Is the ID useful for the user?

migrations/versions/5c2f3eee5f90_introduce_workspaces.py
aponcedeleonch: Agree, thought exactly the same thing over night

src/codegate/pipeline/workspace/workspace.py
aponcedeleonch: If we have the name of the workspace unique then the `id` is not useful. Will remove

src/codegate/db/models.py
```diff
@@ -37,6 +38,12 @@ class Setting(pydantic.BaseModel):
     other_settings: Optional[Any]
 
 
+class Workspace(pydantic.BaseModel):
+    id: Any
+    name: str
```
lukehinds, line 43: Just an idea, I wonder if we may want to use the org: name: format. For example, I could then sort views of repositories via the stacklok org (and see codegate,minder etc)

migrations/versions/5c2f3eee5f90_introduce_workspaces.py
aponcedeleonch: I had an offline discussion with @JAORMX . We agreed to add a `sessions` table that we can re-user later for other purposes if we need to. That approach is also more data-resilient

src/codegate/pipeline/workspace/workspace.py
```diff
@@ -0,0 +1,161 @@
...
+        if args is None or len(args) == 0:
+            return "Please provide a name. Use `codegate-workspace add your_workspace_name`"
+
+        new_workspace_name = args[0]
+        if not new_workspace_name:
+            return "Please provide a name. Use `codegate-workspace add your_workspace_name`"
+
+        workspaces = await self._db_reader.get_workspace_by_name(new_workspace_name)
+        if workspaces:
+            return f"Workspace **{new_workspace_name}** already exists"
```
JAORMX, line 50: wouldn't we know that this results in an error given that the creation would fail because of the `UNIQUE` index? maybe we should try and catch the event to verify if a constraint is violated, and output the friendly error message instead. One less SQL call :smile:

src/codegate/db/connection.py
```diff
@@ -223,26 +232,73 @@ async def record_context(self, context: Optional[PipelineContext]) -> None:
         except Exception as e:
             logger.error(f"Failed to record context: {context}.", error=str(e))
 
+    async def add_workspace(self, workspace_name: str) -> Optional[Workspace]:
```
JAORMX, line 235: Should we do input validation here? That is, let's only allow a subset of characters for workspace names. When chatting, we agreed on alphanumeric with dashes as a good start.

aponcedeleonch: Added it directly into the pydantic model

src/codegate/pipeline/workspace/workspace.py
aponcedeleonch: You're right! Took into account and now that validation is gone :)

JAORMX (approved):
elegant AF

shipit :ship:

Issue #599, opened 2025-01-15
Not detecting malicious npm package


danbarr:
### Describe the issue

In this code snippet, CodeGate is not detecting the malicious `coloramajs` package:

```javascript
const fs = require('fs');
const colorama = require('coloramajs');
```

I tested with both Copilot and Continue, it's missed with both.

### Steps to Reproduce

Add the above snippet to a file and use it as context to a prompt.

### Operating System

MacOS (Arm)

### IDE and Version

VS Code 1.96.3

### Extension and Version

Copilot and Continue

### Provider

GitHub Copilot

### Model

Any

PR #605, opened 2025-01-16
fix: solve some of the problems with snippets and languages

yrobla:
Closes: #599

### DIFF
--- a/src/codegate/pipeline/extract_snippets/extract_snippets.py
+++ b/src/codegate/pipeline/extract_snippets/extract_snippets.py
@@ -98,6 +98,11 @@ def extract_snippets(message: str) -> List[CodeSnippet]:
         # format ` ```python ` in output snippets
         if filename and not matched_language and "." not in filename:
             lang = filename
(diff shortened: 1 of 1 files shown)


src/codegate/pipeline/extract_snippets/extract_snippets.py
```diff
@@ -118,6 +123,9 @@ def extract_snippets(message: str) -> List[CodeSnippet]:
                     if lang not in available_languages:
                         lang = None
 
+        #  just correct the typescript exception
+        if lang == "typescript":
```
lukehinds, line 127: I wonder if it might be worth doing this quickly, as we may have to do more of these mappings:

```python
lang_map = {
    "typescript": "javascript"
}
lang = lang_map.get(lang, lang)
```

lukehinds (approved):

lukehinds (dismissed):
Quick suggestion, but optional if it slows you done.

lukehinds (approved):

PR #611, opened 2025-01-16
Group workflows by trigger type and initial commit for the integration tests workflow

rdimitrov:
The following PR bundles some of the existing workflows by trigger type and adds an initial workflow for the integration tests. 

**Changes:**
* It adds 3 new workflows that run on PR, push to main and published release
* Updates the existing workflows to be triggered by a workflow call and removes their existing triggers
* Updates the README CI badge to refer to the new workflow
* Adds a placeholder for the integration tests (disabled for now). Note that merging this PR would allow to move forward with adding/validating the tests work (needed because one cannot dispatch a workflow if it didn't exist)

**Reasoning:** 
* Treating workflows as functions (known as reusable workflows) and organising them like that makes it more visible what workflows are being run for each trigger type and also allows to easily extend this list in the future. 
* In this case this helps adding the integration tests more easily as it re-uses the container image built from the previous workflow and aims to re-use it in the integration steps by passing it as an artifact.


jhrozek:
this is looking great! Let's figure out the TLS errors tomorrow

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,91 @@
...
+          path: .venv
+          key: venv-${{ runner.os }}-${{ matrix.python-version }}-${{ hashFiles('**/poetry.lock') }}
+
+      - name: Install dependencies
+        run: poetry install --with dev
+
+      - name: Run integration tests
+        env:
+          CODEGATE_PROVIDERS: "copilot"
+          CA_CERT_FILE: "/home/runner/work/codegate/codegate/codegate_volume/certs/ca.crt"
```
lukehinds, line 88: You might be able to fix the ca stuff with something like this:

```bash
cp ca.crt /usr/local/share/ca-certificates/codegate.crt
update-ca-certificates
```

I don't know if sudo would be needed, do these run as root in github?

jhrozek: We should be able to either run the tests without installing the CA_CERT_FILE or not specify the CA_CERT_FILE argument if we had installed the cert to the system store.

.github/workflows/image-build.yml
```diff
@@ -12,15 +16,33 @@ jobs:
   docker-image:
     name: Check docker image build
     runs-on: ubuntu-latest
+    env:
+      IMAGE_NAME: stacklok/codegate
+      IMAGE_TAG: dev
```
jhrozek, line 21: We don't push this image do we? It's just a tag used for the artifact that we upload and then download?

rdimitrov: We don't, no. This is just so it has a reference that we can then use to save it as a tar file.

.github/workflows/image-build.yml
```diff
@@ -12,15 +16,33 @@ jobs:
...
+        id: download-artifact
+        uses: dawidd6/action-download-artifact@80620a5d27ce0ae443b965134db88467fc607b43 # v7
+        with:
+          github_token: ${{ github.token }}
+          workflow: ".github/workflows/import_packages.yml"
+          workflow_conclusion: success
+          name: sqlite_data
+          name_is_regexp: true
+          skip_unpack: false
+          if_no_artifact_found: ignore
```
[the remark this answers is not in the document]
rdimitrov: I got this part from the image-publish workflow to solve one of the issues I faced which was - codegate did not had any populated db so the `invokehttp` test was failing because of that. With this step I'm adding the latest db. 

That said, I don't think this is an expected use case (we call a workflow that does produce an artifact).

rdimitrov:
> @rdimitrov not related to this patch, but iirc you mentioned on Slack that if one of these test cases fails the whole test run wouldn't fail, do we have an issue for that?

I don't think we have, but I should have fixed it as part of this PR (check the changes on `integration_tests.py`). If one or more tests fail now we'll return a non-zero code in oppose to before.

jhrozek:
> > @rdimitrov not related to this patch, but iirc you mentioned on Slack that if one of these test cases fails the whole test run wouldn't fail, do we have an issue for that?
> 
> I don't think we have, but I should have fixed it as part of this PR (check the changes on `integration_tests.py`). If one or more tests fail now we'll return a non-zero code in oppose to before.

Oh, I missed those, I was just reading the workflows. Sloppy review..

.github/workflows/image-build.yml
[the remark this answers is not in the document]
jhrozek: interesting - so the CI images are expected to not have the full db, but the contents of https://github.com/stacklok/codegate/tree/main/data. I worked around that in the initial integration tests patch by adding `invokehttp` into the JSON files. Which is not ideal, if we can copy the db, then much better. I'd just be careful if we are not taking too much time or resources.

jhrozek (dismissed):
thank you for answering my questions during the review!

lukehinds (approved):

lukehinds:
@dependabot rebase

.github/workflows/image-build.yml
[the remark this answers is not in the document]
rdimitrov: We were looking at this issue with Pankaj and Michelangelo and it turns out the dataset was not getting imported while building the image. We discussed that using the latest one vs the minimal dataset should not impose any performance implications (and if it does, it would be nice to catch those, right). 

Another concern against this was if in the tests we rely that a given package is malicious, but it turns out its maintainers improved its score and this is no longer the case. Even though possible, this is rarely going to be the case, so thus the decision.

yrobla (approved):

Issue #555, opened 2025-01-12
Add a simple version check to inform users of outdated versions


lukehinds:
**Description:** 

 Currently, there’s no built-in way for users to know if their instance of CodeGate is running an outdated version. This can cause a sub-par experience or potential security or compatibility issues.

I’d like to propose a simple version-checking system that:

1. **Stores the current local version** (e.g., in a `VERSION` file at the repo root or an environment variable).  
2. **Fetches the latest version** from GitHub (via either:  
   * [Releases API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release), or  
   * The raw `VERSION` file on the `main` branch:  
      \`\`\` [https://raw.githubusercontent.com/](https://raw.githubusercontent.com/)//main/VERSION \`\`\` ).  
3. **Compares** the local version to the latest version.  
4. **Notifies** users (in the frontend or logs) when a newer version is available.

### **Why this is needed**

* **No current version-awareness**: At present, there’s no way of knowing if a user is running an outdated version of CodeGate.  
* **User-friendly**: A quick version check can warn users when they need to update CodeGate.  
* **Lightweight**: Fetching a single file or release tag from GitHub is straightforward and doesn’t require infrastructure.

### **Proposed Steps**

1. **Create/maintain a `VERSION` file** at the root of the repository or using the existing https://github.com/stacklok/codegate/blob/main/src/codegate/__init__.py#L17, and have it updated each time we release new code (automate via GH actions when we publish a release).
2. **Implement a `/api/version` endpoint** in the FastAPI backend that returns the local (running) version.  
3. **Add a check against GitHub** (e.g., `/api/update-status`), fetching the latest release or raw `VERSION` file from the `main` branch.  
4. **Compare** local vs. remote versions and return a response indicating whether the app is up to date (e.g., `"is_latest": true/false`).  
5. **Notify the front end** (React) whenever a mismatch is found, so users can see a banner or message like, “A new version of CodeGate is available\!”

alex-mcgovern:
In implementing this (polling the server from the UI, which checks Github for latest release) I got myself rate limited by the Github API pretty quickly. Going to rethink the polling mechanism in the UI while I work on this.

Issue #579, opened 2025-01-14
Investigate what's up with building codegate with llama-cpp-python 0.3.6


jhrozek:
I don't have more details, but llama-cpp-python 0.3.6 broke our image builds. We just reverted the dep bump without investigating more, but we should to allow us to be up-to-date on the package.

aponcedeleonch:
More context:

Strangely, building the image locally failed using `make image-build` but not on CI with our action `image-build`. The culprit was `llama-cpp-python (0.3.6)`. Logs on failure below.

### Steps to Reproduce

1. Change the version of `llama_cpp_python` in `pyproject.toml` from
```
llama_cpp_python = "==0.3.5"
```
to
```
llama_cpp_python = "==0.3.6"
```
2. Run locally `make image-build`

```logs
#0 25.77   
#0 25.77   -- Configuring done (0.6s)
#0 25.77   -- Generating done (0.0s)
#0 25.77   -- Build files have been written to: /tmp/tmp0d6vwh24/build
#0 25.77   *** Building project with Ninja...
#0 25.77   Change Dir: '/tmp/tmp0d6vwh24/build'
#0 25.77   
#0 25.77   Run Build Command(s): /tmp/tmpy46tdodh/.venv/bin/ninja -v
[...]

PR #612, opened 2025-01-16
feat: version endpoint

alex-mcgovern:
Early work to add a simple version check endpoint to the server, to be polled by the dashboard to address #555.  

Still figuring out how to get stuff done in this codebase — so probably riddled with schoolboy errors.


pyproject.toml
```diff
@@ -40,7 +40,7 @@ build = ">=1.0.0"
 wheel = ">=0.40.0"
 litellm = ">=1.52.11"
 pytest-asyncio = "0.25.2"
-llama_cpp_python = ">=0.3.2"
+llama_cpp_python = "==0.3.5"
```
alex-mcgovern, line 43: There seems to be an issue in 0.3.6 that breaks the image build:

```
#0 25.77   *** CMake build failed
#0 25.77   
#0 25.77 
#0 25.77   at /usr/local/lib/python3.12/site-packages/poetry/installation/chef.py:164 in _prepare
#0 25.77       160│ 
#0 25.77       161│                 error = ChefBuildError("\n\n".join(message_parts))
#0 25.77       162│ 
#0 25.77       163│             if error is not None:
#0 25.77     → 164│                 raise error from None
#0 25.77       165│ 
#0 25.77       166│             return path
#0 25.77       167│ 
#0 25.77       168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:
#0 25.77 
#0 25.77 Note: This error originates from the build backend, and is likely not a problem with poetry but with llama-cpp-python (0.3.6) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "llama-cpp-python (==0.3.6)"'.
```

src/codegate/server.py
```diff
@@ -94,6 +113,26 @@ async def log_user_agent(request: Request, call_next):
...
+            
+            return {
+                "current_version": current_version,
+                "latest_version": latest_version_stripped,
+                "is_latest": is_latest,
+            }
+        except HTTPException as e:
+            return {"current_version": __version__, "latest_version": "unknown", "error": e.detail}
+
+
```
aponcedeleonch, line 135: nit: where is this check meant to be displayed? In the dashboard or somewhere else? If it's meant to be displayed in the dashboard I would move this to the endpoints at `src/codegate/dashboard`. This file `src/codegate/server.py` is more meant for general config, you can add endpoints so if this is the only one of this type.

Also, I would convert the endpoint to `def version_check` instead of `async def version_check`. This is because of the way FastAPI handles the responses, if it's only `def` then it will run in it's own thread, otherwise may get queued up for a little while.

alex-mcgovern: Awesome, thanks for the tip in the second para. Will move it 👍🏻

pyproject.toml
lukehinds: https://github.com/stacklok/codegate/issues/579

src/codegate/dashboard/dashboard.py
```diff
@@ -18,13 +21,30 @@
...
+async def fetch_latest_version(client: AsyncClient) -> str:
+    url = "https://api.github.com/repos/stacklok/codegate/releases/latest"
+    headers = {
+        "Accept": "application/vnd.github+json",
+        "X-GitHub-Api-Version": "2022-11-28"
+    }
+    response = await client.get(url, headers=headers)
+    response.raise_for_status()
+    data = response.json()
+    return data.get("tag_name", "unknown")
```
aponcedeleonch, line 42: More like a curiosity question, why did you decide to use an `AsyncClient` if anyways the wrapper function was going to be non-async. Wouldn't this be a perfect use-case to use `requests`?

src/codegate/dashboard/dashboard.py
```diff
@@ -61,6 +81,43 @@ async def stream_sse():
     """
     return StreamingResponse(generate_sse_events(), media_type="text/event-stream")
 
+@dashboard_router.get("/dashboard/version")
+def version_check(client: AsyncClient = Depends(get_http_client)):
+    try:
+        with ThreadPoolExecutor() as executor:
+            latest_version = executor.submit(fetch_latest_version_sync, client).result()
```
aponcedeleonch, line 88: `def version_check` will already run in it's own thread, executing `fetch_latest_version_sync` seems like an overkill here since also there's nothing else running in parallel.

aponcedeleonch (approved):
2 small questions but I will approve since I tested it and it works which is the main thing :) , nice!

src/codegate/dashboard/dashboard.py
alex-mcgovern: Replaced with requests — much simpler :)

src/codegate/dashboard/dashboard.py
alex-mcgovern: Also cleaned up 👍🏻

aponcedeleonch (approved):
nice

PR #613, opened 2025-01-16
Enhance logging.

blkt:
This change adds line numbers to log messages, and synergizes with Copilot specific changes that make exception handling in some of the copilot pipeline more localized, making it easier to track down issues with the proxy. There's no fundamental change in business logic.

### DIFF
--- a/src/codegate/codegate_logging.py
+++ b/src/codegate/codegate_logging.py
@@ -88,6 +88,7 @@ def setup_logging(
             [
                 structlog.processors.CallsiteParameter.MODULE,
                 structlog.processors.CallsiteParameter.PATHNAME,
+                structlog.processors.CallsiteParameter.LINENO,
             ]
         ),
     ]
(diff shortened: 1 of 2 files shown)


yrobla:
I am open to opinion from other colleagues, but i was very used to set log-level=DEBUG to see litellm outpout, so i could see input/output that is sent to LLMs. With this change i cannot see them anymore. So we either improve our debug logs a bit more to include those outputs, or we should be able to set log levels to litellm to the same level as we pass by parameter

lukehinds:
we should be able to bring the others loggers in , i planned on making it so you could configure what other loggers are allowed in, this way you can switch off stuff like asyncio that is really loud when coupled with sqlalchemy

I had old branch I never completed work on, where I did something like this to make it configurable: 

```python
# Configure logger levels for different components
    for logger_name, level in logger_levels.items():
        logger = logging.getLogger(logger_name)
        logger.setLevel(level)
        logger.propagate = True

    # Default logger levels if not specified
    if "uvicorn" not in logger_levels:
        logging.getLogger("uvicorn").setLevel(log_level.value)
    if "sqlalchemy" not in logger_levels:
        logging.getLogger("sqlalchemy").setLevel(log_level.value)
    if "asyncio" not in logger_levels:
        logging.getLogger("asyncio").setLevel(log_level.value)
```

jhrozek (approved):

PR #614, opened 2025-01-16
If the LLM uses ellipsis in place of the REDACTED string, print the original version

jhrozek:
I was able to make the LLM to reply with `REDACTED=<...>` when
summarizing code. Because our regex and the prefix marker expected the
redacted string to contain the dollar sign, we never flushed the prefix
buffer correctly and ended up later flushing repeatedly.

### DIFF
--- a/src/codegate/pipeline/secrets/secrets.py
+++ b/src/codegate/pipeline/secrets/secrets.py
@@ -312,8 +312,8 @@ class SecretUnredactionStep(OutputPipelineStep):
     """Pipeline step that unredacts protected content in the stream"""
 
     def __init__(self):
-        self.redacted_pattern = re.compile(r"REDACTED<\$([^>]+)>")
-        self.marker_start = "REDACTED<$"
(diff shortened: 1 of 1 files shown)


src/codegate/pipeline/secrets/secrets.py
```diff
@@ -312,8 +312,8 @@ class SecretUnredactionStep(OutputPipelineStep):
     """Pipeline step that unredacts protected content in the stream"""
 
     def __init__(self):
-        self.redacted_pattern = re.compile(r"REDACTED<\$([^>]+)>")
-        self.marker_start = "REDACTED<$"
+        self.redacted_pattern = re.compile(r"REDACTED<(\$?[^>]+)>")
```
JAORMX, line 315: As with all regexes, this will start getting harder to maintain with time. Mind adding a couple of examples expected patterns?

JAORMX (approved):

PR #620, opened 2025-01-16
Initial CRUD API for workspaces

JAORMX:
This adds a simple and unimplemented REST API for workspaces. Workspaces
will be the base for all other resources in terms of REST resource
mapping, so these go first.

These are initially left entirely unimplemented as
https://github.com/stacklok/codegate/pull/600 needs to merge

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -1,0 +1,69 @@
+from fastapi import APIRouter, Response
+from fastapi.exceptions import HTTPException
+from fastapi.routing import APIRoute
+
+from codegate.api import v1_models
+from codegate.pipeline.workspace import commands as wscmd
+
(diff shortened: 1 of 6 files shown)


lukehinds (commented):
I wonder if we should also have the dashboard APIs under api (can be done in a different PR).

Do you think we should introduce api version handling while we have the opportunity and set this as api/v1 ?

aponcedeleonch (dismissed):

aponcedeleonch (approved):

JAORMX:
> Do you think we should introduce api version handling while we have the opportunity and set this as api/v1 ?

I'm already doing this, we actually have `/api/v1`

aponcedeleonch (approved):

PR #634, opened 2025-01-17
Add workspaces to OpenAPI spec

alex-mcgovern:
This might be a little early but moved some stuff around so that we can add the workspaces endpoints to the OpenAPI spec, so we can begin prototyping the UI for this.

### DIFF
--- a/api/openapi.json
+++ b/api/openapi.json
@@ -83,10 +83,217 @@
           }
         }
       }
+    },
+    "/api/v1/workspaces": {
+      "get": {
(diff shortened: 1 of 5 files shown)


JAORMX:
Why is this not automated?

JAORMX (dismissed):

alex-mcgovern:
> Why is this not automated?

Do you mean the generation, or including routers for generation?

JAORMX:
The openapi.json generation, that's right

PR #637, opened 2025-01-17
Add `/api/v1` routes to openapi spec

aponcedeleonch:
It was missing an import

### DIFF
--- a/src/codegate/dashboard/dashboard.py
+++ b/src/codegate/dashboard/dashboard.py
@@ -6,8 +6,9 @@ import requests
 import structlog
 from fastapi import APIRouter, Depends, FastAPI
 from fastapi.responses import StreamingResponse
-from codegate import __version__
 
+from codegate import __version__
+from codegate.api.v1 import v1
 from codegate.dashboard.post_processing import (
     parse_get_alert_conversation,
     parse_messages_in_conversations,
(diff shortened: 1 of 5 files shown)


src/codegate/dashboard/dashboard.py
```diff
@@ -113,6 +115,7 @@ def generate_openapi():
 
     # Include your defined router
     app.include_router(dashboard_router)
+    app.include_router(v1, prefix="/api/v1", tags=["CodeGate API"])
```
JAORMX, line 118: wait, but then this will be double included, since it's already here https://github.com/stacklok/codegate/blob/main/src/codegate/server.py#L102

aponcedeleonch: yes, but we're not generating the openapi spec with the `app` in server. Here is where the generation is happening. Not sure but I think it's done this way to avoid including in the spec all the providers routes, e.g. `/ollama`, `/vllm`.

JAORMX: those don't need to be included if we don't want them to. There are flags for that.

JAORMX: Here you go https://github.com/stacklok/codegate/pull/639

aponcedeleonch: Thank you! Now it's being created from `server.py`

alex-mcgovern:
I think we need to take another look at this, I seem to be getting duplicate entries in the openapi json when I run this locally, and there's also this log line being included in the json file:
```
2025-01-17 13:26:06 [debug    ] Connecting to DB from path: /Users/alexmcgovern/src/codegate/codegate_volume/db/codegate.db
2025-01-17 13:26:06 [debug    ] Connecting to DB from path: /Users/alexmcgovern/src/codegate/codegate_volume/db/codegate.db
```

I think we should remove this debug log here: https://github.com/stacklok/codegate/blob/ec104ce1e2e1f90d6d2d5d2875fe0e086a6b53a2/src/codegate/db/connection.py#L55

alex-mcgovern:
I have prepared an alternative PR where we mount the entire dashboard router under the v1 api router: https://github.com/stacklok/codegate/pull/634

We can then just use the fastAPI app to build the openAPI schema, overall is a lot cleaner I think.

PR #641, opened 2025-01-17
Use exceptions for handling workspace add error

JAORMX:
This stops using the boolean and instead will raise exceptions if
there's an issue adding a workspace. This will help us differentiate if
the operation failed due to a name already being taken, or the name
having invalid characters.

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -1,8 +1,10 @@
 from fastapi import APIRouter, Response
 from fastapi.exceptions import HTTPException
 from fastapi.routing import APIRoute
+from pydantic import ValidationError
 
 from codegate.api import v1_models
+from codegate.db.connection import AlreadyExistsError
 from codegate.workspaces.crud import WorkspaceCrud
 
 v1 = APIRouter()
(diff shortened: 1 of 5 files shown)


src/codegate/pipeline/cli/commands.py
```diff
@@ -69,13 +72,15 @@ async def _add_workspace(self, args: List[str]) -> str:
...
-        if not workspace_created:
-            return (
-                "Something went wrong. Workspace could not be added.\n"
-                "1. Check if the name is alphanumeric and only contains dashes, and underscores.\n"
-                "2. Check if the workspace already exists."
-            )
+        try:
+            _ = await self.workspace_crud.add_workspace(new_workspace_name)
+        except ValidationError as e:
+            return f"Invalid workspace name: {e}"
```
aponcedeleonch, line 78: I think we should we customize the message that gets returned to the user instead of returning them the exception. We should explicitly say that we only accept `-` and `_`.

image

src/codegate/api/v1.py
```diff
@@ -52,13 +54,17 @@ async def activate_workspace(request: v1_models.ActivateWorkspaceRequest, status
...
-    if not created:
-        raise HTTPException(status_code=400, detail="Failed to create workspace")
-
-    return v1_models.Workspace(name=request.name)
+    try:
+        created = await wscrud.add_workspace(request.name)
+    except AlreadyExistsError:
+        raise HTTPException(status_code=409, detail="Workspace already exists")
+    except ValidationError as e:
+        raise HTTPException(status_code=400, detail=str(e))
```
aponcedeleonch, line 62: Probably here it's ok since we control the FE

aponcedeleonch (dismissed):
Left a cosmetic comment but otherwise seems good

aponcedeleonch (approved):

PR #643, opened 2025-01-17
feat: Let the user add their own system prompts

aponcedeleonch:
Related: #454

This PR introduces being able to add system-prompts to workspaces. The workspace system prompts are added in the step that we are using to add CodeGate system prompt. The order of the prompts will be the following assuming all of them are present:
- CodeGate system prompt
- Workspace system prompt
- Request system prompt

The corresponding CLI command is also created and the documentation in the CLI is updated

### DIFF
--- a/src/codegate/db/connection.py
+++ b/src/codegate/db/connection.py
@@ -275,6 +275,20 @@ class DbRecorder(DbCodeGate):
             raise AlreadyExistsError(f"Workspace {workspace_name} already exists.")
         return added_workspace
 
+    async def update_workspace(self, workspace: Workspace) -> Optional[Workspace]:
+        sql = text(
+            """
+            UPDATE workspaces SET
+            name = :name,
+            system_prompt = :system_prompt
+            WHERE id = :id
+            RETURNING *
+            """
+        )
+        # We only pass an object to respect the signature of the function
+        updated_workspace = await self._execute_update_pydantic_model(workspace, sql)
+        return updated_workspace
+
     async def update_session(self, session: Session) -> Optional[Session]:
         sql = text(
             """
@@ -392,11 +406,11 @@ class DbReader(DbCodeGate):
         workspaces = await self._execute_select_pydantic_model(WorkspaceActive, sql)
         return workspaces
 
-    async def get_workspace_by_name(self, name: str) -> List[Workspace]:
+    async def get_workspace_by_name(self, name: str) -> Optional[Workspace]:
         sql = text(
             """
             SELECT
-                id, name
+                id, name, system_prompt
             FROM workspaces
             WHERE name = :name
             """
--- a/src/codegate/db/connection.py
+++ b/src/codegate/db/connection.py
@@ -256,8 +256,7 @@ class DbRecorder(DbCodeGate):
         It may raise a ValidationError if the workspace name is invalid.
         or a AlreadyExistsError if the workspace already exists.
         """
-        workspace = Workspace(id=str(uuid.uuid4()), name=workspace_name)
-
+        workspace = Workspace(id=str(uuid.uuid4()), name=workspace_name, system_prompt=None)
         sql = text(
             """
             INSERT INTO workspaces (id, name)
--- a/src/codegate/db/connection.py
+++ b/src/codegate/db/connection.py
@@ -248,7 +248,7 @@ class DbRecorder(DbCodeGate):
         except Exception as e:
             logger.error(f"Failed to record context: {context}.", error=str(e))
 
-    async def add_workspace(self, workspace_name: str) -> Optional[Workspace]:
+    async def add_workspace(self, workspace_name: str) -> Workspace:
         """Add a new workspace to the DB.
 
         This handles validation and insertion of a new workspace.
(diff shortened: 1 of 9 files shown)


src/codegate/api/v1.py
```diff
@@ -62,9 +62,12 @@ async def create_workspace(request: v1_models.CreateWorkspaceRequest):
...
     except ValidationError:
-        raise HTTPException(status_code=400,
-                            detail=("Invalid workspace name. "
-                                    "Please use only alphanumeric characters and dashes"))
+        raise HTTPException(
+            status_code=400,
+            detail=(
+                "Invalid workspace name. " "Please use only alphanumeric characters and dashes"
+            ),
+        )
```
JAORMX, line 70: can we avoid introducing cosmetic changes or fix our linting job to detect these?

src/codegate/db/connection.py
```diff
@@ -30,9 +30,11 @@
 alert_queue = asyncio.Queue()
 fim_cache = FimCache()
 
+
```
JAORMX, line 33: cosmetic

src/codegate/db/connection.py
```diff
@@ -30,9 +30,11 @@
...
 class AlreadyExistsError(Exception):
     pass
 
+
```
JAORMX, line 37: cosmetic

src/codegate/db/connection.py
```diff
@@ -266,12 +268,27 @@ async def add_workspace(self, workspace_name: str) -> Optional[Workspace]:
 
         try:
             added_workspace = await self._execute_update_pydantic_model(
-                workspace, sql, should_raise=True)
+                workspace, sql, should_raise=True
+            )
```
JAORMX, line 272: cosmetic

src/codegate/pipeline/cli/commands.py
```diff
@@ -31,10 +31,10 @@ async def run(self, args: List[str]) -> str:
     @property
     def help(self) -> str:
         return (
-            "### CodeGate Version\n\n"
+            "### CodeGate Version\n"
             "Prints the version of CodeGate.\n\n"
+            "*args*: None\n\n"
```
JAORMX, line 36: This seems unrelated to this change.

src/codegate/pipeline/cli/commands.py
```diff
@@ -66,41 +67,64 @@ async def _add_workspace(self, args: List[str]) -> str:
...
         if not was_activated:
             return (
-                f"Workspace **{workspace_name}** does not exist or was already active. "
+                f"Workspace `{workspace_name}` does not exist or was already active. "
                 f"Use `codegate workspace add {workspace_name}` to add it"
             )
-        return f"Workspace **{workspace_name}** has been activated"
+        return f"Workspace `{workspace_name}` has been activated"
+
+    async def _add_system_prompt(self, args: List[str]):
```
JAORMX, line 106: since this does an update let's rename it as such.

src/codegate/workspaces/crud.py
```diff
@@ -77,3 +78,20 @@ async def activate_workspace(self, workspace_name: str) -> bool:
         db_recorder = DbRecorder()
         await db_recorder.update_session(session)
         return True
+
+    async def update_workspace_system_prompt(
+        self, workspace_name: str, sys_prompt_lst: List[str]
+    ) -> Optional[Workspace]:
+        selected_workspace = await self._db_reader.get_workspace_by_name(workspace_name)
```
JAORMX, line 85: can we throw an exception instead of returning `None`?

src/codegate/pipeline/cli/commands.py
```diff
@@ -66,41 +67,64 @@ async def _add_workspace(self, args: List[str]) -> str:
...
+            workspace_name, sys_prompt_lst
+        )
+        if not updated_worksapce:
+            return (
+                f"Workspace system prompt not updated. "
+                f"Check if the workspace `{workspace_name}` exists"
+            )
+        return (
+            f"Workspace `{updated_worksapce.name}` system prompt "
+            f"updated to:\n```\n{updated_worksapce.system_prompt}\n```"
```
JAORMX, line 126: The prompt could be quite long, should we just say "successfully updated" instead?

src/codegate/pipeline/cli/commands.py
```diff
@@ -46,6 +46,7 @@ def __init__(self):
             "list": self._list_workspaces,
             "add": self._add_workspace,
             "activate": self._activate_workspace,
+            "system-prompt": self._add_system_prompt,
```
JAORMX, line 49: should we add a command to list the current prompt?

JAORMX (changes requested):

src/codegate/workspaces/crud.py
aponcedeleonch: Yes, nice suggestion. Done so

src/codegate/pipeline/cli/commands.py
aponcedeleonch: Yes, nice idea. Done

src/codegate/pipeline/cli/commands.py
aponcedeleonch: Yes, I think with the `show` system-prompt command now is redundant

src/codegate/pipeline/cli/commands.py
aponcedeleonch: You're right. I thought to do some formatting along with this PR but let's keep it simple

JAORMX (approved):
shipit :ship:

PR #649, opened 2025-01-20
Polish errors and miscelaneous fixes for workspaces

JAORMX:
This raises specific errors which are then handled excplicitly. Thus
giving us more detailed error messages.


aponcedeleonch (approved):
Nice!

PR #657, opened 2025-01-20
Implement basic soft deletion for workspaces

JAORMX:
this adds a `deleted_at` column to workspaces that implements a basic
soft-deletion mechanism. All relevant queries have been modified to reflect this.

At the moment, there is no hard deletion of workspaces; this will be
implemented in the future.

We also have no way of showing "archived" or "soft-deleted" workspaces.
This will come in due time.

### DIFF
--- a/src/codegate/pipeline/cli/commands.py
+++ b/src/codegate/pipeline/cli/commands.py
@@ -212,7 +212,7 @@ class Workspace(CodegateCommandSubcommand):
             return "An error occurred while activating the workspace"
         return f"Workspace **{workspace_name}** has been activated"
 
-    async def _remove_workspace(self, args: List[str]) -> str:
+    async def _remove_workspace(self, flags: Dict[str, str], args: List[str]) -> str:
         """
         Remove a workspace
         """
--- a/src/codegate/pipeline/cli/commands.py
+++ b/src/codegate/pipeline/cli/commands.py
@@ -153,6 +153,7 @@ class Workspace(CodegateCommandSubcommand):
             "list": self._list_workspaces,
             "add": self._add_workspace,
             "activate": self._activate_workspace,
+            "remove": self._remove_workspace,
         }
 
     async def _list_workspaces(self, flags: Dict[str, str], args: List[str]) -> str:
@@ -210,6 +211,25 @@ class Workspace(CodegateCommandSubcommand):
         except Exception:
             return "An error occurred while activating the workspace"
         return f"Workspace **{workspace_name}** has been activated"
+
+    async def _remove_workspace(self, args: List[str]) -> str:
+        """
+        Remove a workspace
+        """
+        if args is None or len(args) == 0:
+            return "Please provide a name. Use `codegate workspace remove workspace_name`"
+
+        workspace_name = args[0]
+        if not workspace_name:
+            return "Please provide a name. Use `codegate workspace remove workspace_name`"
+
+        try:
+            await self.workspace_crud.soft_delete_workspace(workspace_name)
+        except crud.WorkspaceDoesNotExistError:
+            return f"Workspace **{workspace_name}** does not exist"
+        except Exception:
+            return "An error occurred while removing the workspace"
+        return f"Workspace **{workspace_name}** has been removed"
 
     @property
     def help(self) -> str:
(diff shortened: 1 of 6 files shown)


src/codegate/pipeline/cli/commands.py
```diff
@@ -153,6 +153,7 @@ def subcommands(self) -> Dict[str, Callable[[List[str]], Awaitable[str]]]:
             "list": self._list_workspaces,
             "add": self._add_workspace,
             "activate": self._activate_workspace,
+            "remove": self._remove_workspace,
         }
 
     async def _list_workspaces(self, flags: Dict[str, str], args: List[str]) -> str:
```
aponcedeleonch, line 159: We need to change the `_list_workspaces` to return only the workspaces that haven't been deleted. As it is it will return all workspaces which wouldn't be correct.

src/codegate/workspaces/crud.py
```diff
@@ -100,6 +96,31 @@ async def update_workspace_system_prompt(
...
+            raise WorkspaceCrudError("Cannot delete default workspace.")
+
+        selected_workspace = await self._db_reader.get_workspace_by_name(workspace_name)
+        if not selected_workspace:
+            raise WorkspaceDoesNotExistError(f"Workspace {workspace_name} does not exist.")
+
+        # Check if workspace is active, if it is, make the default workspace active
+        active_workspace = await self._db_reader.get_active_workspace()
+        if active_workspace and active_workspace.id == selected_workspace.id:
+            _ = await self.activate_workspace("default")
```
aponcedeleonch, line 115: I would prefer to not automatically change for the user to the "default" workspace. What about if they made a mistake and they don't actually want to delete the active workspace. It will serve kind of like a double-step confirmation. Specially since we don't currently have a way of restoring the workspaces.

Not critical, I just think it's good to have a discussion about it

aponcedeleonch (changes requested):
Couple of comments. The one that I would like to get attended is the one about listing the workspaces

src/codegate/pipeline/cli/commands.py
JAORMX: We only list non-deleted workspaces since the SQL query itself was deleted. We don't need to change the logic in `_list_workspaces`.

aponcedeleonch (approved):

src/codegate/workspaces/crud.py
JAORMX: OK, I can return an error instaed.

src/codegate/pipeline/cli/commands.py
aponcedeleonch: Right, see it now. Sorry for my blindness

aponcedeleonch (changes requested):
Needs a fix

src/codegate/pipeline/cli/commands.py
```diff
@@ -211,6 +212,25 @@ async def _activate_workspace(self, flags: Dict[str, str], args: List[str]) -> s
             return "An error occurred while activating the workspace"
         return f"Workspace **{workspace_name}** has been activated"
 
+    async def _remove_workspace(self, args: List[str]) -> str:
```
aponcedeleonch: I changed the arguments that all the functions get. It should be now
```
async def _remove_workspace(self, flags: Dict[str, str] args: List[str]) -> str:
```
Note the extra `flags` param

JAORMX: Good catch! I had missed this in the rebase

src/codegate/pipeline/cli/commands.py
```diff
@@ -211,6 +212,25 @@ async def _activate_workspace(self, flags: Dict[str, str], args: List[str]) -> s
...
+        workspace_name = args[0]
+        if not workspace_name:
+            return "Please provide a name. Use `codegate workspace remove workspace_name`"
+
+        try:
+            await self.workspace_crud.soft_delete_workspace(workspace_name)
+        except crud.WorkspaceDoesNotExistError:
+            return f"Workspace **{workspace_name}** does not exist"
+        except Exception:
+            return "An error occurred while removing the workspace"
```
aponcedeleonch, line 231: We should give the user a clearer error message of what's going on. I got this when trying to delete the active workspace. It's ok to fix on a later PR.

aponcedeleonch (approved):

src/codegate/pipeline/cli/commands.py
JAORMX: Can you open a bug about this? I can fix it tomorrow.

PR #664, opened 2025-01-20
Add utilities to dump Copilot traffic.

blkt:
This change adds a utility routine to dump raw data tapped from the wire to multiple files in a temporary directory under the path specified via `CODEGATE_DUMP_DIR`.

No instrumentation is performed if `CODEGATE_DUMP_DIR` is `None`. The temporary directory is not deleted when the process exits.

NOTE: the utility does not ensure the folder specified via `CODEGATE_DUMP_DIR` is writable, which might cause user-visible failures at startup.

### DIFF
--- a/src/codegate/providers/copilot/provider.py
+++ b/src/codegate/providers/copilot/provider.py
@@ -1,6 +1,10 @@
 import asyncio
+import contextlib
+import datetime
+import os
 import re
 import ssl
+import tempfile
 from dataclasses import dataclass
 from typing import Dict, List, Optional, Tuple, Union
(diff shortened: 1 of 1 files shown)


jhrozek:
@blkt great, thank you, can you take this out of draft so we can ack and merge? I'll add a similar utility to the cursor provider, just an inline copy for now, and once we refactor we can merge the two.

src/codegate/providers/copilot/provider.py
```diff
@@ -1,6 +1,10 @@
 import asyncio
+import contextlib
```
jhrozek, line 2: do we use this import?

Issue #671, opened 2025-01-21
Add ability to rename a workspace


JAORMX:
This applies to both the API and the pseudo-CLI

PR #683, opened 2025-01-21
Add constructs to rename workspaces

JAORMX:
This adds the CRUD API endpoints and pseudo-CLI subcommands to rename
workspaces. Note that `default` and `active` are reserved keywords and
cannot be used.

Closes: https://github.com/stacklok/codegate/issues/671

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -61,8 +61,18 @@ async def activate_workspace(request: v1_models.ActivateWorkspaceRequest, status
 
 
 @v1.post("/workspaces", tags=["Workspaces"], generate_unique_id_function=uniq_name, status_code=201)
-async def create_workspace(request: v1_models.CreateWorkspaceRequest) -> v1_models.Workspace:
+async def create_workspace(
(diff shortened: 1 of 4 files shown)


src/codegate/api/v1.py
```diff
@@ -83,6 +93,30 @@ async def create_workspace(request: v1_models.CreateWorkspaceRequest) -> v1_mode
     return v1_models.Workspace(name=request.name, is_active=False)
 
 
+async def rename_workspace(
+    request: v1_models.CreateOrRenameWorkspaceRequest,
+) -> v1_models.Workspace:
+    try:
+        _ = await wscrud.rename_workspace(request.name, request.rename_to)
+    except crud.WorkspaceDoesNotExistError:
```
peppescg, line 101: do we need to add a case for avoiding rename the default one?

peppescg (approved):
🚀

src/codegate/api/v1.py
JAORMX: This is handled in src/codegate/workspaces/crud.py

Issue #669, opened 2025-01-21
Hard-delete workspaces


JAORMX:
After doing https://github.com/stacklok/codegate/issues/668 we should also add the ability to finally hard-delete a workspace.

That is, we nee to add the `ON DELETE CASCADE` constructs in the database for each foreign key that references a workspace.

We should also have a CRUD method to hard-delete an archived workspace.

This should encompass both the REST API and the pseudo-CLI.

aponcedeleonch:
Closed by: https://github.com/stacklok/codegate/pull/686

PR #685, opened 2025-01-21
Hard delete workspaces

aponcedeleonch:
Closes: https://github.com/stacklok/codegate/issues/669

To hard delete a workspace it first needs to be soft deleted.

### DIFF
--- a/src/codegate/db/connection.py
+++ b/src/codegate/db/connection.py
@@ -8,7 +8,8 @@ import structlog
 from alembic import command as alembic_command
 from alembic.config import Config as AlembicConfig
 from pydantic import BaseModel
-from sqlalchemy import CursorResult, TextClause, text
+from sqlalchemy import CursorResult, TextClause, event, text
+from sqlalchemy.engine import Engine
 from sqlalchemy.exc import IntegrityError, OperationalError
 from sqlalchemy.ext.asyncio import create_async_engine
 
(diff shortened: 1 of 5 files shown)


peppescg (approved):

src/codegate/db/connection.py
```diff
@@ -431,7 +460,7 @@ async def get_workspaces(self) -> List[WorkspaceActive]:
         workspaces = await self._execute_select_pydantic_model(WorkspaceActive, sql)
         return workspaces
 
-    async def get_workspace_by_name(self, name: str) -> Optional[Workspace]:
+    async def get_non_deleted_workspace_by_name(self, name: str) -> Optional[Workspace]:
```
JAORMX, line 463: most of the time we'd be dealing with non deleted workspaces. I would suggest keeping the name `get_workspace_by_name` since it adheres to the expectation.

src/codegate/db/connection.py
```diff
@@ -446,6 +475,21 @@ async def get_workspace_by_name(self, name: str) -> Optional[Workspace]:
         )
         return workspaces[0] if workspaces else None
 
+    async def get_workspace_by_name(self, name: str) -> Optional[Workspace]:
```
JAORMX, line 478: should we rename this to deal with the expectation that we're dealing with soft deleted or archived workspaces?

JAORMX (changes requested):

JAORMX:
superceded by https://github.com/stacklok/codegate/pull/686

Issue #700, opened 2025-01-22
Adjust README to remove redundancy and differences with Development Guide


wright-io:
README.md has some details in the "Development" section which are redundant with, and sometimes different from the guidance in the Development Guide (docs/development.md). The README currently points to the Development Guide.

The proposed approach would be:
- Retain the link to the developer reference guides in the README
- Remove the detailed development information from the README (moving anything unique and useful into the development guide if it isn't already there).

PR #701, opened 2025-01-22
Harmonize readme and dev guide (fixes https://github.com/stacklok/codegate/issues/700)     

wright-io:


docs/development.md
```diff
@@ -38,7 +38,14 @@ from potential AI-related security risks. Key features include:
...
 
-3. Install project dependencies:
+3. Set up virtual environment
+
+   ```bash
+   python -m venv venv
+   source venv/bin/activate  # On Windows: venv\Scripts\activate
+   ```
+
+4. Install project dependencies:
```
lukehinds, line 48: Thanks for the PR Doug, looks good. FYI.. don't need to create a venv with poetry, it creates its own when you run poetry install, you can then just run poetry shell to activate the env


image

wright-io: @lukehinds Makes sense! Would you prefer we advise just using `poetry run` in the docs for simplicity (as opposed to bringing in the shell plugin option to those docs)? For example:

> Then run import_packages to ensure integration test data is created: 
> ```bash
> poetry run python scripts/import_packages.py
> ```
>
> Then the integration tests can be executed by running:
> ```bash
> python tests/integration/integration_tests.py
> ```

wright-io:
@lukehinds Would appreciate another review on this, if you don't mind.

lukehinds:
Yeah, don't worry about mentioning the shell plugin.

lukehinds (approved):

PR #712, opened 2025-01-22
Fix `greenlet` dependency in poetry lock file.

blkt:
It looks like poetry on macOS does not install `greenlet`, while it does on Linux, both for version 1.8.4 and 2.0.1, which seems to be a bug. Additionally, failures only manifest after _using_ codegate, and not during startup.

This change adds `greenlet` package to `pyproject.toml`.


pyproject.toml
```diff
@@ -29,6 +29,7 @@ tree-sitter-rust = "==0.23.2"
 alembic = "==1.14.1"
 pygments = "==2.19.1"
 sqlite-vec-sl-tmp = "==0.0.4"
+greenlet = "^3.1.1"
```
JAORMX, line 32: can we pin this dependency?

lukehinds (dismissed):

lukehinds:
@dependabot rebase

JAORMX (approved):

aponcedeleonch (approved):

Issue #726, opened 2025-01-22
Workspace name case sensitivity inconsistent


danbarr:
### Describe the issue

I'm seeing some inconsistencies with how case is handled in workspace names. I created a workspace with a name in ALL_CAPS, and it was converted to lower case:

<img width="426" alt="Image" src="https://github.com/user-attachments/assets/d9adea00-62dc-4196-a463-959158f9bb41" />

That's fine, we can document that workspace names are always lower case. But then subsequent commands using the same ALL_CAPS don't work:

<img width="426" alt="Image" src="https://github.com/user-attachments/assets/2d3c9702-00a9-422e-8a28-c17a9fc40f56" />

I think we should be consistent in case (in)sensitivity.

### Steps to Reproduce

From a chat prompt, run:

```plain
codegate workspace add WORKSPACE_NAME

# See that workspace is created as workspace_name:
codegate workspace list

# Try commands using mixed case:
codegate workspace activate WORKSPACE_NAME
codegate workspace archive Workspace_Name
```

### Operating System

MacOS (Arm)

### IDE and Version

N/A

### Extension and Version

N/A

### Provider

GitHub Copilot

### Model

N/A

### Codegate version

v0.1.12

PR #736, opened 2025-01-23
Allow workspace operations to work without caring about the case

JAORMX:
Ultimately all operations will work regardless of the case. Though the
final storage format will be lowercase.

This also made changes to the db models to disambiguate the names from one another.

Closes: https://github.com/stacklok/codegate/issues/726

### DIFF
--- a/src/codegate/db/models.py
+++ b/src/codegate/db/models.py
@@ -30,7 +30,7 @@ class Prompt(BaseModel):
     workspace_id: Optional[str]
 
 
-WorskpaceNameStr = Annotated[
+WorkspaceNameStr = Annotated[
     str,
     StringConstraints(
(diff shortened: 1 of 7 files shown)


src/codegate/db/models.py
```diff
@@ -30,20 +30,43 @@ class Prompt(BaseModel):
...
 
 
-WorskpaceNameStr = Annotated[
+class Setting(BaseModel):
+    id: Any
+    ip: Optional[Any]
+    port: Optional[Any]
+    llm_model: Optional[Any]
+    system_prompt: Optional[Any]
+    other_settings: Optional[Any]
```
aponcedeleonch, line 39: This was deleted recently. We no longer have a DB table called `settings`

aponcedeleonch (approved):

aponcedeleonch (approved):

Issue #732, opened 2025-01-22
Copilot replying the workspace already exists for new workspaces


aponcedeleonch:
### Describe the issue

Copilot replying a new workspace already exists. See image below. 

<img width="275" alt="Image" src="https://github.com/user-attachments/assets/01be578b-0c4a-433a-b5d2-910587973e3f" />

We indeed get >2 requests. For example, if we add a workspace with `codegate workspace add foobar` in the first request we add correctly the workspace and reply appropriately but for some reason the client doesn't get the answer. Then on the second request the workspace is already created and we answer with `Workspace foobar already exists` (which is true since it was created by the first request).

### Steps to Reproduce

Go into copilot and add a new workspace
```
codegate workspace add foobar
```

### Operating System

MacOS (Intel)

### IDE and Version

1.96.3

### Extension and Version

Copilot 1.257.0

### Provider

GitHub Copilot

### Model

GPT-4

### Codegate version

v0.1.13

### Additional Context

Request number 1. We create the Workspace and respond appropriately:
```json
{
  "messages": [
    {
      "role": "system",
[...]

PR #738, opened 2025-01-23
Adds cache to cli-chat commands

aponcedeleonch:
Closes: #732

Copilot sends at least 2 requests with the same `last_user_message`. Hence we execute the same command 2 times and reply to the last one. The last behaviour will cause that if we create a Workspace with Copilot the cli will respond that the workspace already exists.

This PR implements a workaround to cache the commands and it's outputs. That way we can reply the same to subsequent requests sent by Copilot


src/codegate/pipeline/cli/commands.py
```diff
@@ -16,6 +17,15 @@ class NoSubcommandError(Exception):
     pass
 
 
+class ExecCommand(BaseModel):
+
+    cmd_out: str
+    exec_time: datetime.datetime
+
+
+command_cache: Dict[str, ExecCommand] = {}
```
JAORMX, line 26: What if you instead of a TTLCache from cachetools? That providers automatic expiration which will keep the cache's size in check in the long run. This solution works, but the lack of cleanup will result in increased memory usage in the long run. What do you think?

aponcedeleonch: I like the suggestion. Implemented

JAORMX (approved):

PR #743, opened 2025-01-23
Enable the integration tests for Anthropic

rdimitrov:
Fixes: #742

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -12,6 +12,9 @@ on:
       copilot-key:
         description: 'The Copilot key to use for integration tests'
         required: true
+      anthropic-key:
+        description: 'The Anthropic key to use for integration tests'
+        required: true
 
 jobs:
   integration-tests:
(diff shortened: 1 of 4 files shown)


tests/integration/testcases.yaml
```diff
@@ -217,15 +217,7 @@ testcases:
...
     likes: |
-      Hello! I'm CodeGate, your security-focused AI assistant. I can help you with:
-
-      - Software security analysis and reviews
-      - Package security assessments
-      - Secure coding practices and guidance
-      - Security vulnerability analysis
-      - Best practices for secure implementations
-
-      How can I assist you with your security needs today?
```
rdimitrov, line 228: I had to change this otherwise it was failing the test. I made it similar to the test we did for Copilot. 

Of course if we expect to have a codegate-aware answer let me know as we might have a bug 👍

jhrozek: no, I think this was a remnant of us adding the codegate prompt to everything

.github/workflows/integration-tests.yml
```diff
@@ -129,6 +132,13 @@ jobs:
         run: |
           poetry run python tests/integration/integration_tests.py
 
+      - name: Run integration tests - Anthropic
+        env:
+          CODEGATE_PROVIDERS: "anthropic"
+          ENV_ANTHROPIC_KEY: ${{ secrets.anthropic-key }}
```
jhrozek, line 138: we might want to do a matrix in the future, but this is good enough

jhrozek (dismissed):

.github/workflows/integration-tests.yml
rdimitrov: I was thinking about that too, but I wanted to get everything running first and then do that as a follow up 👍 That way they can run in parallel too

jhrozek (approved):

Issue #420, opened 2024-12-19
Secrets discovered provide a link to the main conversation


lukehinds:
When Secrets are discovered, the UI provides a link to the main conversation, with the secrets highlighted in red.

PR #744, opened 2025-01-23
Properly parse conversations into alerts

aponcedeleonch:
Closes: #420

This PR introduces:
- Have full conversations in alerts. Until now we were only showing the message where the alert came from in `*/alerts`
- Match the `chat_id` in the alert conversation to the ones returned in `*/messages`
- Fixes a grouping issue in which conversations with secrets were not grouped correctly because of the `REDACTED<>` text.


JAORMX (approved):

JAORMX (approved):

peppescg (approved):
Tested it locally and it works well 🚀

PR #756, opened 2025-01-24
make `type` in PartialQuestions more strict

kantord:


src/codegate/api/v1_models.py
```diff
@@ -87,6 +88,11 @@ class QuestionAnswer(pydantic.BaseModel):
     answer: Optional[ChatMessage]
 
 
+class PartialQuestionsType(str, Enum):
```
aponcedeleonch, line 91: I would make the name of the Enum more generic like `QuestionsType` and also apply it to the `Conversation` object further down the file

kantord: makes sense

kantord: pushed

aponcedeleonch (approved):

PR #757, opened 2025-01-24
fix: solve problems with codegate cli and previous context

yrobla:
Cline was not properly detecting the codegate cli commands when there was some previous context

Also fix a problem with Copilot and cli commands


aponcedeleonch (changes requested):
For some reason, the commands are not working for me on Continue. The following command should list the help:
image

Correct output for reference:
image

yrobla:
ok will take a look

> For some reason, the commands are not working for me on Continue. The following command should list the help: image
> 
> Correct output for reference: image

ok will take a look

aponcedeleonch (approved):

PR #764, opened 2025-01-24
nit: replace delete with archive in workspace crud errors

alex-mcgovern:
Just makes the error messages for soft-deleting workspaces a little bit more consistent with surrounding copy in the frontend, (We render them directly on screen).


peppescg (approved):
🚀

JAORMX (approved):

Issue #765, opened 2025-01-24
Adding utm_source=codegate not working


wright-io:
### Describe the issue

https://github.com/stacklok/codegate/pull/749 made changes to add utm_source=codegate to insight links. Although the copilot integration tests passed, the querystring is not actually appearing in the generated links through copilot. 

It appears additional changes are needed to make this work.

It also looks like the integration test approach might need to be adjusted. I'll create a separate issue for this. 

### Steps to Reproduce

- Run the latest container in dev
- Configure codegate with copilot 
- Ask copilot "Help me use invokehttp"
- Receive CodeGate warning about invokehttp, but without the utm_source=codegate querystring

### Operating System

MacOS (Arm)

### IDE and Version

VS Code 1.96.4

### Extension and Version

Copilot 1.258.0

### Provider

GitHub Copilot

### Model

GPT-4o

### Codegate version

v0.1.13-25c2368-dev

PR #766, opened 2025-01-24
Fix issue with utm_source not appearing in generated insight links

wright-io:
Closes https://github.com/stacklok/codegate/issues/765


wright-io:
@lukehinds @yrobla I know there is still a lot of manual testing going on. I verified this change locally with the image now for copilot, but we may want to regression test in the other codegen tools. Tests pass and the behavior is as expected, but I don't know for sure if we are making assumptions about the insight link formats anywhere that isn't explicitly tested.

JAORMX (approved):

PR #780, opened 2025-01-27
Add mocked token-usage endpoints

aponcedeleonch:
For the moment mocked the token usage endpoints. This would be the request with most of the information
```bash
curl 'http://localhost:8989/api/v1/workspaces/default/token-usage'
```

It will bring the total tokens used in the workspace. Additionally it will have a breakdown of how many tokens were used by each model in the workspace


peppescg (approved):
LGTM

Issue #418, opened 2024-12-19
Display Token usage


lukehinds:
Can we display the amount of tokens used by any given provider, this would be useful for the new copilot free tier. 

An extra would be to record the token usage per conversation. This would allow user insight into what prompts are more costly and allow optimization. 

kudos @craigmcl for the idea. 

<img width="709" alt="Image" src="https://github.com/user-attachments/assets/64a8624e-426a-443a-b010-735d336c7bb6" />

lukehinds:
this will require #454 to land first, so let's keep in backlog for now.

aponcedeleonch:
At an initial investigation the used tokens are not listed neither in the request nor the response from the LLM.

Request
```json
{
  "messages": [...],
  "model": "gpt-4o",
  "temperature": 0.1,
  "top_p": 1,
  "max_tokens": 4096,
  "n": 1,
  "stream": true
}
```

`max_tokens` are the maximum tokens accepted not the **used** tokens.

Response
```text
[
"{\"id\":\"\",\"created\":0,\"model\":\"\",\"object\":\"chat.completion.chunk\",\"choices\":[]}", 
"{\"id\":\"chatcmpl-Ao5A9Sf7Q6WB751oF5OpU7Wmwcfv4\",\"created\":1736499609,\"model\":\"gpt-4o-2024-05-13\",\"object\":\"chat.completion.chunk\",\"choices\":[{\"index\":0,\"delta\":{\"content\":\"\",\"role\":\"assistant\"}}]}", 
....
"{\"id\":\"chatcmpl-Ao5A9Sf7Q6WB751oF5OpU7Wmwcfv4\",\"created\":1736499609,\"model\":\"gpt-4o-2024-05-13\",\"object\":\"chat.completion.chunk\",\"choices\":[{\"finish_reason\":\"stop\",\"index\":0,\"delta\":{\"role\":\"assistant\"}}]}"]
```

There are 2 alternatives:
1. See if there's a way the LLM providers list in their response the tokens they have used. At a first glance it looks to be possible at least for [OpeanAI](https://help.openai.com/en/articles/6614209-how-do-i-check-my-token-usage)
2. Use our own tokenizer. We could tokenize ourselves the request and response and calculate that way the number of used tokens. The big drawback with this is that the tokens we calculate with the tokenizer may not match the tokens used by the LLM. But at least it would be an approximation

aponcedeleonch:
[...]

PR #788, opened 2025-01-27
Include the token usage for every conversation and workspace

aponcedeleonch:
Closes: #418

This PR does introduces the changes necessary to track the used tokens per request, process them, and to return them in the API.

Specific changes:
- Make sure we process all the stream and record to DB at the very end
- Include the flag `"stream_options": {"include_usage": True},` so the providers respond with the tokens
- Added the necessary processing for the API
- Modified the initial API models to display correctly the tokens and its price

Notable:
- Ollama and Llama CPP tokens cannot be obtained with the current code. It's not critical since these are local providers and shouldn't represent any cost in tokens
  - Ollama: If we switch our requests to not use the Python client but directly HTTP requests we would be able to obtain the tokens
  - Llama CPP: I couldn't find a way to get the tokens when streaming the response.
- Copilot is not returning the used tokens. Also not critical since Copilot charges a fixed price and not per token

Sample response
```json
curl -Ss 'http://localhost:8989/api/v1/workspaces/tokens/token-usage' | jq
{
  "tokens_by_model": {
    "hosted_vllm/unsloth/Qwen2.5-Coder-32B-Instruct": {
      "provider_type": "vllm",
      "model": "hosted_vllm/unsloth/Qwen2.5-Coder-32B-Instruct",
      "token_usage": {
        "input_tokens": 30,
        "output_tokens": 10,
        "input_cost": 0.0,
        "output_cost": 0.0
      }
    },
    "gpt-4o-mini": {
      "provider_type": "openai",
      "model": "gpt-4o-mini",
      "token_usage": {
        "input_tokens": 8,
        "output_tokens": 10,
        "input_cost": 0.0000012,
        "output_cost": 0.000006
      }
    },
    "claude-3-5-sonnet-latest": {
      "provider_type": "anthropic",
      "model": "claude-3-5-sonnet-latest",
      "token_usage": {
        "input_tokens": 471,
        "output_tokens": 35,
        "input_cost": 0.001413,
        "output_cost": 0.000525
      }
    }
  },
  "token_usage": {
    "input_tokens": 509,
    "output_tokens": 55,
    "input_cost": 0.0014142,
    "output_cost": 0.000531
  }
}
```


src/codegate/db/token_usage.py
```diff
@@ -0,0 +1,90 @@
...
+logger = structlog.get_logger("codegate")
+
+# 1 day cache. Not keep all the models in the cache. Just the ones we have used recently.
+model_cost_cache = TTLCache(maxsize=2000, ttl=1 * 24 * 60 * 60)
+
+
+class TokenUsageParser:
+
+    def __init__(self):
+        self.url = "https://raw.githubusercontent.com/BerriAI/litellm/main/model_prices_and_context_window.json"
```
JAORMX, line 20: This is fine for a workaround but it's a little funky to depend on an external JSON file for runtime. I would suggest having a CI job download that file periodically, embed it in a file in the project here, and subsequently read that from codegate. It's a more resilient solution.

aponcedeleonch: nice idea, I added the file to the `/data` folder and I also added a GH workflow to get the file and update it every week

JAORMX (dismissed):

JAORMX (approved):

Issue #785, opened 2025-01-27
Implement model discovery for provider endpoints


JAORMX:
When registering a provider endpoint, we should be able to discover what models does the provider support.

Issue #784, opened 2025-01-27
Implement crud manager for provider endpoints


JAORMX:
This will implement the business logic of interfacing with the database and configuration for provider endpoints.

PR #790, opened 2025-01-28
Kick off provider endpoint CRUD structure and registration

JAORMX:
This structure will handle all the database operations and turn that
into the right models.

Note that for provider endpoints we already have a way of setting these
via configuration, so this is taken into account to output some sample
objects that users can leverage.

Each provider will need to implement a `models` function which allows us
to auto-discover models for a provider.

Closes: #784
Closes: #785 

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -1,21 +1,24 @@
 from typing import List, Optional
+from uuid import UUID
 
 import requests
 import structlog
-from fastapi import APIRouter, HTTPException, Response
+from fastapi import APIRouter, Depends, HTTPException, Response
 from fastapi.responses import StreamingResponse
 from fastapi.routing import APIRoute
-from pydantic import ValidationError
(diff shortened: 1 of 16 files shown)


src/codegate/api/v1.py
```diff
@@ -25,38 +28,78 @@ def uniq_name(route: APIRoute):
...
+    return [provend]
+
+
+# This needs to be above /provider-endpoints/{provider_id} to avoid conflict
+@v1.get(
+    "/provider-endpoints/models",
+    tags=["Providers"],
+    generate_unique_id_function=uniq_name,
+)
+async def list_all_models_for_all_providers() -> List[v1_models.ModelByProvider]:
```
aponcedeleonch, line 62: Would be good to also accept here `FilterByNameParams` to be able to get the models of a specific provider

aponcedeleonch (approved):

PR #802, opened 2025-01-29
Enable the integration tests for Ollama

rdimitrov:
The following PR enables the integration tests for Ollama.

**Details:**
* Starts the CodeGate container in the host network (so everything can talk to each other without the need to specify explict ports), we don't have port conflicts anyway
* Starts the Ollama container and tests if it runs
* Downloads and runs the `qwen2.5-coder:0.5b` model
* Updates the tests to use `qwen2.5-coder:0.5b`
* Updates the chat test to expect a simpler welcome message
* Adds a missing header in the tests

Fixes: https://github.com/stacklok/codegate/issues/815


jhrozek:
I think this is a bug in the text that uncovers a bug in codegate:
```
decoded_line = ' {"error": "[Errno -2] Name or service not known", "type": "error",      │ │
│ │                    "choices": ['+2                                                           │ │
│ │                e = IndexError('list index out of range')                                     │ │
│ │        json_line = {                                                                         │ │
│ │                    │   'error': '[Errno -2] Name or service not known',                      │ │
│ │                    │   'type': 'error',                                                      │ │
│ │                    │   'choices': []                                                         │ │
│ │                    }                                                                         │ │
│ │             line = b'data: {"error": "[Errno -2] Name or service not known", "type":         │ │
│ │                    "error", "choice'+7                                                       │ │
```

So what happens here is that we get a line and expect it to cointain `choices` but it only contains an `error`. We should handle the error, perhaps by raising an HTTP exception. But the root cause is that we cannot reach a service (or ollama can't?)

rdimitrov:
> I think this is a bug in the text that uncovers a bug in codegate:
> 
> ```
> decoded_line = ' {"error": "[Errno -2] Name or service not known", "type": "error",      │ │
> │ │                    "choices": ['+2                                                           │ │
> │ │                e = IndexError('list index out of range')                                     │ │
> │ │        json_line = {                                                                         │ │
> │ │                    │   'error': '[Errno -2] Name or service not known',                      │ │
> │ │                    │   'type': 'error',                                                      │ │
> │ │                    │   'choices': []                                                         │ │
> │ │                    }                                                                         │ │
> │ │             line = b'data: {"error": "[Errno -2] Name or service not known", "type":         │ │
> │ │                    "error", "choice'+7                                                       │ │
> ```
> 
> So what happens here is that we get a line and expect it to cointain `choices` but it only contains an `error`. We should handle the error, perhaps by raising an HTTP exception. But the root cause is that we cannot reach a service (or ollama can't?)

Thanks! I'll file a separate issue for this and fix it there.

jhrozek (approved):
yay!

PR #826, opened 2025-01-29
Bootstrap provider models on addition & implement mux endpoints

JAORMX:
When adding a provider via the API, this bootstrap the available models
for a provider.

### DIFF
--- a/src/codegate/providers/crud/crud.py
+++ b/src/codegate/providers/crud/crud.py
@@ -11,7 +11,7 @@ from codegate.config import Config
 from codegate.db import models as dbmodels
 from codegate.db.connection import DbReader, DbRecorder
 from codegate.providers.base import BaseProvider
-from codegate.providers.registry import ProviderRegistry
+from codegate.providers.registry import ProviderRegistry, get_provider_registry
 
 logger = structlog.get_logger("codegate")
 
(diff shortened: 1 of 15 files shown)


src/codegate/providers/crud/crud.py
```diff
@@ -62,22 +62,60 @@ async def get_endpoint_by_name(self, name: str) -> Optional[apimodelsv1.Provider
...
     async def update_endpoint(
         self, endpoint: apimodelsv1.ProviderEndpoint
     ) -> apimodelsv1.ProviderEndpoint:
         """Update an endpoint."""
 
+        founddbe = await self._db_reader.get_provider_endpoint_by_id(endpoint.id)
+        if founddbe is None:
+            raise ProviderNotFoundError("Provider not found")
+
+        # TODO: We should probably allow this and reinitialize the provider
```
aponcedeleonch, line 115: Agree with this TODO

src/codegate/providers/llamacpp/provider.py
```diff
@@ -27,9 +28,15 @@ def __init__(
...
+    def models(self, api_key: str = None) -> List[str]:
+        headers = {}
+        if api_key:
+            headers["Authorization"] = f"Bearer {api_key}"
         # HACK: This is using OpenAI's /v1/models endpoint to get the list of models
-        resp = httpx.get(f"{self.base_url}/v1/models")
+        resp = httpx.get(
+            f"{self.base_url}/v1/models",
+            headers=headers,
+        )
```
aponcedeleonch, line 39: Normally `llamacpp` models are simple `.gguf` files. I don't think there's a request to a server we can make. What we could do is check which models we have in our `codegate_volume/models` folder and list them here. We have documentation saying that's the correct way of adding your `llamacpp` models to CodeGate

src/codegate/providers/openai/provider.py
```diff
@@ -35,9 +35,12 @@ def __init__(
...
 
-    def models(self) -> List[str]:
-        # NOTE: This won't work since we need an API Key being set.
-        resp = httpx.get(f"{self.lm_studio_url}/v1/models")
+    def models(self, api_key: str = None) -> List[str]:
+        headers = {}
+        if api_key:
+            headers["Authorization"] = f"Bearer {api_key}"
+
+        resp = httpx.get(f"{self.lm_studio_url}/v1/models", headers=headers)
```
aponcedeleonch, line 43: What if we actually want to talk to OpenAI, this is assuming LM Studio. Maybe we need another provider that is called LM Studio to handle this correctly?

src/codegate/providers/crud/crud.py
```diff
@@ -62,22 +62,60 @@ async def get_endpoint_by_name(self, name: str) -> Optional[apimodelsv1.Provider
         return apimodelsv1.ProviderEndpoint.from_db_model(dbendpoint)
 
     async def add_endpoint(
-        self, endpoint: apimodelsv1.ProviderEndpoint
+        self, endpoint: apimodelsv1.AddProviderEndpointRequest
```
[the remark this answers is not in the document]
JAORMX: Not anymore, I did create a dedicated API class so we can pass additional POST-specific parameters.

src/codegate/providers/openai/provider.py
JAORMX: fixed.

Issue #803, opened 2025-01-29
Implement "Synthetic" provider


JAORMX:
This provider will automatically detect which override is enabled for a workspace and forward to the right place. We'll be using this instead of the per-provider implementations we used to have.

PR #827, opened 2025-01-29
Create mux endpoint and route to appropriate destination provider

aponcedeleonch:
Closes: #803 

This PR introduces the following changes related with muxing:
- Introduces a new route to which the muxing requests will go to: `/v1/mux`
- When hitting the above route the following steps will be performed
  1. The format of the request will be recognized. For the moment is only relevant to determine if the request has an Antropic body or an OpenAI body
  2. Get the destination provider from the active workspace configuration
  3. Transform the incoming request to the format of the destination provider
  4. Route the request to the adequate destination provider
- Introduction of a new method in all providers: `process_request`. The purpose of the method is to take the raw data, api_key and url path of the request and handle them appropriately. This is needed to support step number 4. above

### DIFF
--- a/src/codegate/db/models.py
+++ b/src/codegate/db/models.py
@@ -1,4 +1,5 @@
 import datetime
+from enum import Enum
 from typing import Annotated, Any, Dict, Optional
 
 from pydantic import BaseModel, StringConstraints
@@ -116,6 +117,19 @@ class Session(BaseModel):
 # Models for select queries
 
 
+class ProviderType(str, Enum):
+    """
+    Represents the different types of providers we support.
+    """
+
+    openai = "openai"
+    anthropic = "anthropic"
+    vllm = "vllm"
+    ollama = "ollama"
+    lm_studio = "lm_studio"
+    llamacpp = "llamacpp"
+
+
 class GetPromptWithOutputsRow(BaseModel):
     id: Any
     timestamp: Any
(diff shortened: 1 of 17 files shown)


src/codegate/providers/anthropic/provider.py
```diff
@@ -44,6 +44,22 @@ def models(self) -> List[str]:
 
         return [model["id"] for model in respjson.get("data", [])]
 
+    async def process_request(self, data: dict, api_key: str, request_url_path: str):
```
jhrozek, line 47: I think this is a nice change

src/codegate/mux/adapter.py
```diff
@@ -0,0 +1,65 @@
+import copy
+
+from codegate.db import models as db_models
+
+
+class BodyAdapter:
```
jhrozek, line 6: This might conflict/be duplicated with what @blkt is doing

aponcedeleonch: Yes, I had a chat with him. We will replace this implementation with his once it's ready

src/codegate/db/models.py
```diff
@@ -116,6 +117,19 @@ class Session(BaseModel):
...
+    """
+    Represents the different types of providers we support.
+    """
+
+    openai = "openai"
+    anthropic = "anthropic"
+    vllm = "vllm"
+    ollama = "ollama"
+    lm_studio = "lm_studio"
+    llamacpp = "llamacpp"
```
JAORMX, line 130: nit: these probably shouldn't be db models and should be part of the API instead. Makes more logical sense.

JAORMX (approved):

src/codegate/db/models.py
aponcedeleonch: Kind of agree. I like to have them here so that Pydantic can validate automatically the `models` that use this enum. This can be useful when creating or updating the DB models.

Having them here also avoids circular imports. Maybe we should have a higher abstraction, .e.g. `general_models` that can be used both by DB models and API models to avoid circular imports

JAORMX: I like your proposal better, the higher abstraction

Issue #820, opened 2025-01-29
Codegate not properly picking latest messages from Open Interpreter


yrobla:
It uses a new role "tool" that was not considered by our engine when extracting messages, so it was not properly picking the latest messages for analysis

PR #829, opened 2025-01-29
fix: consider the tool role when being in open interpreter

yrobla:
It uses a new message role tool, that contains some context for using internal tools and the data it provides. So we need to consider it when parsing messages, or when inserting context messages

Closes: #820


src/codegate/pipeline/codegate_context_retriever/codegate.py
```diff
@@ -149,6 +149,14 @@ async def process(
                         # Combine updated task content with the rest of the message
                         context_msg = updated_task_content + rest_of_message
 
+                elif base_tool == "open interpreter":
```
jhrozek, line 152: I'm sorry I'm quite dumb today, could you give me a before and after example of the conversations?

yrobla: i actually changed the logic as it was not working in all cases

yrobla: i refactored this code as the logic was not working for all the cases

src/codegate/pipeline/base.py
```diff
@@ -260,19 +261,25 @@ def get_last_user_message_block(
         messages = request["messages"]
         block_start_index = None
 
+        base_tool = get_tool_name_from_messages(request)
+        accepted_roles = ["user", "assistant"]
+        if base_tool == "open interpreter":
+            # open interpreter also uses the role "tool"
+            accepted_roles.append("tool")
```
jhrozek, line 268: sorry to be difficult about this review.

Would it break anything if we accepted the tools either way, meaning for all clients?

What do we miss if we don't process the tool messages here?

src/codegate/pipeline/base.py
```diff
@@ -260,19 +261,25 @@ def get_last_user_message_block(
...
                     continue
 
-                if messages[i]["role"] == "user":
+                if messages[i]["role"] in ["user", "tool"]:
                     user_messages.append(content_str)
                     block_start_index = i
 
                 # Specifically for Aider, when "Ok." block is found, stop
-                if content_str == "Ok." and messages[i]["role"] == "assistant":
+                if base_tool == "aider" and content_str == "Ok." and messages[i]["role"] == "assistant":
```
jhrozek: how do we detect aider? I looked into get_tool_name_from_messages and I don't see aider returned from there

yrobla: i actually removed that, it is outdated, i removed the check for aider

src/codegate/pipeline/base.py
yrobla: well i was not aware if any other client using tools but wanted to be conservative about it, so that's why i filter by the client. Do you prefer that i open that and we do not filter?

jhrozek (approved):
We did a bunch of testing offline with Yolanda and the code seems to work. I'm going to file a follow-up to support tools with openai in a more general way

PR #832, opened 2025-01-29
Add client detection

jhrozek:
- **Add client detector interface** - Adds a decorator that can be added to the FastAPI handlers and detect the client from a fallback mechanism, by the user-agent, by a specific header or by a matching word in the messages. At the moment, the clients are represented as a simple enum, but in follow-up patches they will be represented by classes that can perform the changes by an interface providing callbacks from the pipeline or other places that need client-specific behaviour.
- **Use the client type when streaming the data to the client, not when executing completion** - We used to special-case ollama stream generation by passing the client type to the execute_completion. Instead, let's pass the client type to the place that needs special casing using the recently introduce client type enum.
- **Use the client type when instantiating and running provider pipelines** - Instead of detecting the client type again when the pipeline is being processed, let's pass the client type on instantiating the pipeline instance as a constant and replace the hardcoded client strings by just using the constants.
- **Remove get_tool_name_from_messages** - This was superseded by using the client enum.
- **Remove the is_copilot flag in favor of using the autodetected client** - In the copilot provider, we can hardcode the client type to copilot when instantiating the pipelines.

### DIFF
--- a/src/codegate/clients/detector.py
+++ b/src/codegate/clients/detector.py
@@ -1,0 +1,224 @@
+import re
+from abc import ABC, abstractmethod
+from functools import wraps
+from typing import List, Optional
+
+import structlog
+from fastapi import Request
+
+from codegate.clients.clients import ClientType
+
+logger = structlog.get_logger("codegate")
+
+
+class HeaderDetector:
+    """
+    Base utility class for header-based detection
+    """
+
+    def __init__(self, header_name: str, header_value: Optional[str] = None):
+        self.header_name = header_name
+        self.header_value = header_value
+
+    def detect(self, request: Request) -> bool:
+        logger.debug(
+            "checking header detection",
+            header_name=self.header_name,
+            header_value=self.header_value,
+            request_headers=dict(request.headers),
+        )
+        # Check if the header is present, if not we didn't detect the client
+        if self.header_name not in request.headers:
+            return False
+        # now we know that the header is present, if we don't care about the value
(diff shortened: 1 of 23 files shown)


src/codegate/clients/detector.py
```diff
@@ -0,0 +1,185 @@
...
+        logger.debug(
+            "checking header detection",
+            header_name=self.header_name,
+            header_value=self.header_value,
+            request_headers=dict(request.headers),
+        )
+        if self.header_name not in request.headers:
+            return False
+        if self.header_value is None:
+            return True
```
aponcedeleonch, line 33: Just double checking, if the value of a header is `None` we return `True` meaning we mark the header as detected. Is that correct? Intuitively I would think we return `False`

aponcedeleonch (dismissed):
Amazing!

src/codegate/clients/detector.py
jhrozek: Yes, this was actually on purpose, but was not understandable from the code. I added some comments, thanks for calling this out.

jhrozek:
I rebased the branch and fixed some smallish issues along the way. Ready for a next round.

jhrozek:
Good guy integration test, catching my bugs.

lukehinds:
Good work @jhrozek , just needs a rebase.

lukehinds (dismissed):

rdimitrov (approved):

PR #837, opened 2025-01-30
Add LM Studio provider

aponcedeleonch:
We were using OpenAI provider to interface with LM Studio since both of them were very similar. For muxing we need to clearly distinguish to which providers we need to route the request. Hence it will be easier to disambiguate the providers.

### DIFF
--- a/src/codegate/providers/lm_studio/provider.py
+++ b/src/codegate/providers/lm_studio/provider.py
@@ -51,8 +51,6 @@ class LmStudioProvider(OpenAIProvider):
             body = await request.body()
             data = json.loads(body)
 
-            # if model starts with lm_studio, propagate it
-            if data.get("model", "").startswith("lm_studio"):
-                data["base_url"] = self.lm_studio_url + "/v1/"
+            data["base_url"] = self.lm_studio_url + "/v1/"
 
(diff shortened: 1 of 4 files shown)


JAORMX (approved):

src/codegate/providers/lm_studio/provider.py
```diff
@@ -0,0 +1,58 @@
...
+            if not authorization.startswith("Bearer "):
+                raise HTTPException(status_code=401, detail="Invalid authorization header")
+
+            api_key = authorization.split(" ")[1]
+            body = await request.body()
+            data = json.loads(body)
+
+            # if model starts with lm_studio, propagate it
+            if data.get("model", "").startswith("lm_studio"):
+                data["base_url"] = self.lm_studio_url + "/v1/"
```
yrobla, line 56: this was used because the way to discriminate if it was a query for openai or for lm studio was via the model name. But if we have a different route, i do not think that we need to have this conditional now, the base url will always be lm_studio_url

aponcedeleonch: I will change it :)

yrobla (approved):

PR #845, opened 2025-01-30
fix: open intepreter not properly working with ollama

yrobla:
It needs fixes in cli commands


aponcedeleonch (approved):
The changes look good but I don't have open interpreter to test. Continue and Copilot are working just fine

PR #848, opened 2025-01-30
Fix the Ollama FIM completion

jhrozek:
We were not passing the suffix at all which was breaking the FIM
completion.

Fixes: #853


JAORMX (approved):

rdimitrov (approved):
Nice!

PR #851, opened 2025-01-30
Implement muxing routing cache

### LINKED ISSUE #786: Implement "catch-all" mux matcher

JAORMX:
This allows us to have a parsed and in-memory representation of the
routing rule engine. The intent is to reduce calls to the database.

This is more expensive to maintain since we need to refresh the cache on
every operation pertaining to models, endpoints, workspaces, and muxes themselves.

Finally, this implements the interface for matchers; Currently we only have one
"catch all" matcher, but this will change in the future.

Closes #786 

Signed-off-by: Juan Antonio Osorio <ozz@stacklok.com> and Alejandro Ponce de León <aponcedeleonch@stacklok.com>

### DIFF
--- a/src/codegate/muxing/rulematcher.py
+++ b/src/codegate/muxing/rulematcher.py
@@ -1,0 +1,118 @@
+import copy
+from abc import ABC, abstractmethod
+from collections import UserDict
+from threading import Lock, RLock
+from typing import List, Optional
+
+from codegate.db import models as db_models
+
+_muxrules_sgtn = None
+
+_singleton_lock = Lock()
+
+
+def get_muxing_rules_registry():
+    """Returns a singleton instance of the muxing rules registry."""
+
+    global _muxrules_sgtn
+
+    if _muxrules_sgtn is None:
+        with _singleton_lock:
+            if _muxrules_sgtn is None:
+                _muxrules_sgtn = MuxingRulesinWorkspaces()
+
+    return _muxrules_sgtn
+
+
+class ModelRoute:
+    """A route for a model."""
+
+    def __init__(
+        self,
+        model: db_models.ProviderModel,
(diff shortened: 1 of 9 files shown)


src/codegate/muxing/rulematcher.py
```diff
@@ -0,0 +1,118 @@
+import copy
+from abc import ABC, abstractmethod
+from collections import UserDict
+from threading import Lock, RLock
```
blkt, line 4: **concern:** I think we should only use locking primitives from the asyncio package, but I might be wrong.
https://docs.python.org/3/library/asyncio-sync.html

aponcedeleonch: Because we use asyncio in other places or what advantages do you see in it? According to the docs
> asyncio primitives are not thread-safe, therefore they should not be used for OS thread synchronization (use [threading](https://docs.python.org/3/library/threading.html#module-threading) for that);

So I think primitives from `threading` are a better choice, but to be honest also not 100% sure. Can be persuaded

blkt: My understanding is that we're using `asyncio` everywhere, so we're single threaded (but still concurrent).

Using `threading` risks locking the only thread of execution and halt the application.

I might be wrong here, but that's my understanding of how async execution environment work in general, not only in Python.

aponcedeleonch: You're right. Do you see it very critical or could we fix it in a later PR? I just saw and making the change would hit some parts of the code which aren't really part of this one.

aponcedeleonch: Nevermind. I made the change, now we're using the primitives from `asyncio`

jhrozek (approved):

PR #866, opened 2025-01-31
Fix python 3.13 support.

blkt:
Python 3.13 dropped module `cgi`, which was made available as a stand-alone library named `legacy-cgi`. A recent bump in `litellm` added it as a dependency, breaking their support for python 3.13.

This change adds the stand-alone version of `cgi` library.

Source https://docs.python.org/3/library/cgi.html


kantord (dismissed):
LGTM

aponcedeleonch (approved):

PR #870, opened 2025-01-31
fix: add specific condition for Kodu in system prompt

yrobla:
It is needed to stop the flow when talking about malicious packages or secrets leaked, otherwise kodu keeps iterating as it is not interpreting our warnings.


jhrozek (approved):

jhrozek:
We need to add this as one of the things for the client API though, this is just a temp fix

jhrozek (approved):

Issue #838, opened 2025-01-30
[Task]: Optimise the integration tests to run in parallel and in separate jobs


rdimitrov:
### Description

The following task is to optimise the integration tests so they:
* Run in parallel instead of sequentially
* Run in separate jobs
* Perhaps add retry mechanism if needed?

This will help make them run faster and also re-run only the one that fails without re-running the rest (saving up tokens).

PR #887, opened 2025-02-03
Run the integration tests in parallel

rdimitrov:
The following PR updates the integration tests workflow so all provider tests run in parallel along with some small tweaks. Overall this should help make the integration tests faster to complete and reduce token quota when retrying.

Fixes: #838


aponcedeleonch (approved):
Amazing

jhrozek (approved):

Issue #904, opened 2025-02-04
[Task]: Add malicious package tests for all providers


rdimitrov:
### Description

The following issue is about adding malicious package tests for all providers, i.e. testing if invokehttp is flagged correctly

rdimitrov:
All of the related PRs are now merged, closing as completed.

PR #912, opened 2025-02-04
Add malicious package test for vllm

rdimitrov:
The following PR adds malicious package test for vllm.

Related to: #904

### DIFF
--- a/tests/integration/vllm/testcases.yaml
+++ b/tests/integration/vllm/testcases.yaml
@@ -63,3 +63,26 @@ testcases:
       if __name__ == '__main__':
           test_call_api()
 
+  vllm_malicious_package_question:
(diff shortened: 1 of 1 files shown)


aponcedeleonch (dismissed):

rdimitrov:
I had to update the malicious package test case for this and the rest of the local providers because of the capabilities of the 0.5B model we are using for these tests. It looks like it cannot handle the system prompts as precisely as we expect which results in non-deterministic responses, i.e. sometimes it includes example code snippets even though it says the package is malicious, sometimes it doesn't include the link to insights. To handle all of this decided to simplify the test and just check if CodeGate was able to flag it.

jhrozek (approved):
the functionality is shared across the providers, so I think this is good enough

Issue #878, opened 2025-02-02
CodeGate doesn't work with either LiteLLM or OpenRouter


tan-yong-sheng:
I am trying to use codegate with cline, and connect LLM api via LiteLLM proxy (hosted on cloud)

Environment: 
- Windows 11

Here are the steps I did:

1. On laptop, I run a docker container for codegate: 

`docker run --name codegate -d -p 8989:8989 -p 9090:9090 -e CODEGATE_OPENAI_URL=https://<MY_LITELLM_URL>/v1 --restart unless-stopped ghcr.io/stacklok/codegate`

2. Then, I tried to set up the credentials for cline in VS code, as follows:

![Image](https://github.com/user-attachments/assets/79c23041-3ab1-4941-a69e-033f7799be45)

I thought as I set the environment variable to CODEGATE_OPENAI_URL=https://litellm.tanyongsheng.site/v1, it will change the default environment variable for openai url

However, I still get this error:

![Image](https://github.com/user-attachments/assets/15e199af-b0f3-4d02-8dce-8ae132cf99d2)

And when I tried to switch to gemini 2.0 that I set up in litellm proxy. There is still error coming up:

![Image](https://github.com/user-attachments/assets/82c2720c-b3b2-4350-8382-29f9e6984d78)

Hope for help, thanks

tan-yong-sheng:
But, it seems like leading to another error, which it doesn't trigger when there are secrets inside the file:

Here is the same credentials in my `config.ini` file

![Image](https://github.com/user-attachments/assets/669ec6ea-392d-49c1-afc7-8c9d3fd30188)

And when I check codegate via docker logs, and I found a lot of errors there:

![Image](https://github.com/user-attachments/assets/2849b795-288c-46c5-aa09-28c47d2be9a5)

And not sure if it's related to the errors above, I found there are some 304 errors for litellm UI? But, I think I didn't access my litellm UI that time.... (Not that sure)

![Image](https://github.com/user-attachments/assets/39f7bc50-f0b0-4202-be17-5e202b8ff55e)

I am thinking it's better for me to try to configure it via openai API first, and then transfer to litellm API ...

jhrozek:
@tan-yong-sheng thanks for filing the issue!

The rate limits were a codegate bug that was fixed in main but not released yet. We'll get a release out shortly.

[...]

PR #921, opened 2025-02-04
Add an OpenRouter provider

jhrozek:
- **Move _get_base_url to the base provider**
- **Add an openrouter provider** - OpenRouter is a "muxing provider" which itself provides access to multiple models and providers. It speaks a dialect of the OpenAI protocol, but for our purposes, we can say it's OpenAI.

There are some differences in handling the requests, though:
1) we need to know where to forward the request to, by default this is `https://openrouter.ai/api/v1`, this is done by setting the base_url parameter
2) we need to prefix the model with `openrouter/`. This is a lite-LLM-ism (see https://docs.litellm.ai/docs/providers/openrouter) which we'll be able to remove once we ditch litellm

Initially I was considering just exposing the OpenAI provider on an
additional route and handling the prefix based on the route, but I think
having an explicit provider class is better as it allows us to handle
any differences in OpenRouter dialect easily in the future.

Initially I only tested this with Cline, using DeepSeek for Plan and Anthropic
for Act. I still need to test other assistants (Continue).

Related: #878

### DIFF
--- a/src/codegate/server.py
+++ b/src/codegate/server.py
@@ -18,6 +18,7 @@ from codegate.providers.llamacpp.provider import LlamaCppProvider
 from codegate.providers.lm_studio.provider import LmStudioProvider
 from codegate.providers.ollama.provider import OllamaProvider
 from codegate.providers.openai.provider import OpenAIProvider
+from codegate.providers.openrouter.provider import OpenRouterProvider
 from codegate.providers.registry import ProviderRegistry, get_provider_registry
 from codegate.providers.vllm.provider import VLLMProvider
 
@@ -74,6 +75,10 @@ def init_app(pipeline_factory: PipelineFactory) -> CodeGateServer:
     registry.add_provider(
         ProviderType.openai,
         OpenAIProvider(pipeline_factory),
+    )
+    registry.add_provider(
+        ProviderType.openai,
+        OpenRouterProvider(pipeline_factory),
     )
     registry.add_provider(
         ProviderType.anthropic,
--- a/src/codegate/server.py
+++ b/src/codegate/server.py
@@ -73,11 +73,11 @@ def init_app(pipeline_factory: PipelineFactory) -> CodeGateServer:
 
     # Register all known providers
     registry.add_provider(
-        ProviderType.openrouter,
+        ProviderType.openai,
         OpenAIProvider(pipeline_factory),
     )
     registry.add_provider(
-        ProviderType.openai,
+        ProviderType.openrouter,
         OpenRouterProvider(pipeline_factory),
     )
     registry.add_provider(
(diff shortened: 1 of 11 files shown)


jhrozek:
Oh I also need to add some tests.

src/codegate/server.py
```diff
@@ -75,6 +76,10 @@ async def log_user_agent(request: Request, call_next):
         ProviderType.openai,
         OpenAIProvider(pipeline_factory),
     )
+    registry.add_provider(
+        ProviderType.openai,
```
jhrozek, line 80: @aponcedeleonch what purpose does the ProviderType have? Since openrouter speaks openai, but fronts many models, how would muxing work with it?

jhrozek:
This was my Cline config
Screenshot 2025-02-04 at 22 27 01

jhrozek:
@danbarr when and if this is accepted, we'll probably want to document how to use OR with this provider. I noticed we suggest to use VLLM now in the docs for continue.

danbarr:
Yeah vllm worked because both it and OpenRouter are an OpenAI-compatible API. Originally I thought we could use our existing /openai provider endpoint but that's where we ran into the LiteLLM automatic routing based on the model name.

src/codegate/server.py
aponcedeleonch: `ProviderType` is an enum we use to pick the correct provider for muxing. See [here](https://github.com/stacklok/codegate/blob/main/src/codegate/muxing/router.py#L72). We look into the `registry` and pick the destination provider from there. This line would override `ProviderType.openai` in the `registry` with the new openrouter provider. It's better to create a new entry in the enum, e.g. `ProviderType.openrouter` and use it here. That way a new entry in the `registry` will get created. Otherwise we would be trying to mux to OpenRouter both openai requests and OpenRouter requests.

For the latter part, on how would muxing work with it, basically it will work as any other provider. Openrouter will just have a ton more of models

jhrozek: ok, I added a new provider as a separate patch for easier review

jhrozek:
> Oh I also need to add some tests.

tests added

src/codegate/server.py
aponcedeleonch: thanks :)

src/codegate/server.py
```diff
@@ -72,9 +73,13 @@ async def log_user_agent(request: Request, call_next):
...
     # Register all known providers
     registry.add_provider(
-        ProviderType.openai,
+        ProviderType.openrouter,
         OpenAIProvider(pipeline_factory),
     )
+    registry.add_provider(
+        ProviderType.openai,
+        OpenRouterProvider(pipeline_factory),
+    )
```
aponcedeleonch, line 82: These are exchanged. Meaning It should be `ProviderType.openai` on the top and `ProviderType.openrouter` on the bottom

aponcedeleonch: no problem :)

aponcedeleonch (approved):

Issue #209, opened 2024-12-05
Assess fidelity of secret pattern matching


poppysec:
A considerable number of the patterns in `signatures.yml` will match on secret variable names, such as `OPENAI_API_KEY` in addition to or in place of the secret key or token itself. This is an artefact of the previous secret blocking implementation. 
```yaml
- Amazon:
  - Access Key: (?:A3T[A-Z0-9]|AKIA|AGPA|AIDA|AROA|AIPA|ANPA|ANVA|ASIA|ABIA|ACCA)[A-Z0-9]{16}
  - Secret Access Key Variable: (?i)(amazon|amz|aws)[-_]{0,1}(secret)[-_]{0,1}((access)[-_]{0,1}){0,1}key
  # - Cognito User Pool ID: (?i)us-[a-z]{2,}-[a-z]{4,}-\d{1,}
  - RDS Password: (?i)(rds\-master\-password|db\-password)
  - S3 Private Key Variable: (?i)AWS_S3_PRIVATE_KEY|s3_key|S3_PRIVATE_KEY
  - Security Token Header Variable: (?i)X-Amz-Security-Token
  - API Gateway Key Source Header Variable: (?i)x-amazon-apigateway-api-key-source
  - S3 Bucket: (?i)AWS_S3_BUCKET|s3_bucket
  - SNS Confirmation URL: (?i)https:\/\/sns\.[a-z0-9-]+\.amazonaws\.com\/?Action=ConfirmSubscription&Token=[a-zA-Z0-9-=_]+
  - SES SMTP Password Variable: (?i)ses_smtp_password
  - AWS Private Key Variable: (?i)ec2\-private\-key|EC2_PRIVATE_KEY
  - MWS Token: (amzn\.mws\.[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})
  - AppSync GraphQL Key: \bda2-[a-z0-9]{26}

- Microsoft:
  - Azure API Key Variable: (?i)Ocp-Apim-Subscription-Key
  - Azure Functions Key Header Variable: (?i)x-functions-key
```
Now with on-the-fly encryption we must be precise with the strings which are encrypted - if we obfuscate an entire line in a user's code prompt, including the variable name, it could cause the LLM to produce mangled output. We also want to avoid adding spurious claims of encrypting x amount of nonexistent secrets to the response.

[...]

PR #937, opened 2025-02-05
feat: add better functionality to secret detection

yrobla:
add the following functionality:
- improve regexes for identificating secrets via patterns
- add high entropy secret detection
- include the key of the secret in the found matches

Closes: #209


prompts/default.yaml
```diff
@@ -13,11 +13,13 @@ default_chat: |
...
 
-  If you are Kodu, also append that when finding malicious packages or leaked secrets:
+  Also append that when finding malicious packages or leaked secrets:
 
-  "** critical vulnerability found, you must take action **
-  Please notify the user with ask follow up tool about the following vulnerability:
-  <vulnerability_content>"
+  "** critical vulnerability found, you must take action **"
+
+  If you are Kodu, also append:
```
rdimitrov, line 20: Should this be part of the prompt? I've been looking at our tests and I think it's acting strange with smaller models, i.e. it includes the `If you are Kodu` part that follows without testing it with a client like that. Would it work if this part of the prompt is again a template that we update based on the client before attaching it to the request?

signatures.yaml
```diff
@@ -265,13 +273,6 @@
 - Postgresql:
     - URL: (?i)(?:pgsql:|postgres:|postgresql:)//[\S]{1,256}:[\S]{1,256}@[-.%\w\/:]+\.[\S]+
 
-- GitHub:
```
rdimitrov, line 268: Sorry for asking but is this intentional? I suppose so seeing you also removed the GitHub test case below. In any case shouldn't we still have tests that cover github's use cases?

prompts/default.yaml
yrobla: i added this as an advice from Kodu maintainers, to avoid multiple loops on Kodu. But is not acting as expected, works partially. I'd like to actually only add it in the case of tool being Kodu

signatures.yaml
yrobla: yes, the problem was that due to a previous bug, all the tokens from github were hardcoded, not picked from the yaml file. When i removed that hardcoding and started to use the regexes from the yaml, i needed to modify tests, and also remove a useless one that was testing the hardcoded behaviour

src/codegate/pipeline/secrets/signatures.py
```diff
@@ -177,25 +189,17 @@ def _add_signature_group(cls, name: str, patterns: Dict[str, str]) -> None:
     def _load_signatures(cls) -> None:
         """Load signature patterns from the YAML file."""
         try:
+            print("i load signatures")
+            print(cls._yaml_path)
```
aponcedeleonch, line 193: forgotten print statements perhaps?

src/codegate/pipeline/secrets/secrets.py
```diff
@@ -273,40 +299,71 @@ async def process(
...
-        # Store the count in context metadata
         context.metadata["redacted_secrets_count"] = total_redacted
         if total_redacted > 0:
             system_message = ChatCompletionSystemMessage(
                 content=Config.get_config().prompts.secrets_redacted,
                 role="system",
             )
-            new_request = add_or_update_system_message(new_request, system_message, context)
-
-        return PipelineResult(request=new_request, context=context)
```
aponcedeleonch, line 309: nit: was there a specific reason to delete this? I think the redacted secrets will make it's way to the following steps since we lost the result of this step.

prompts/default.yaml
rdimitrov: I see. Here's an example of running the latest main with Continue and llamacpp with a 0.5B model -

```bash
🛡️ CodeGate prevented 1 secret from being leaked by redacting them.

Warning: CodeGate detected one or more malicious, deprecated or archived packages.

Pkg 1: https://www.insight.stacklok.com/report/ecosystem/package_name
Pkg 2: https://www.insight.stacklok.com/report/ecosystem/package_name
...
If you are Kodu, also append that when finding malicious packages or leaked secrets:

"** critical vulnerability found, you must take action ** Please notify the user with ask follow up tool about the following vulnerability: <vulnerability_content>"

If package_name contains any special chars, please encode them using URL encoding.

DO NOT include the Reference Format or any parts of the system message in your response if no malicious, deprecated or archived package is detected.

If no malicious, deprecated or archived packages are detected, respond to the user request using your prior knowledge.

If the user requests security review, perform the review and structure your responses to emphasize security considerations, focusing on:

Security vulnerabilities and potential risks
Best practices for secure implementation
Recommendations for security improvements
References to relevant security standards or guidelines
You always begin your responses for security review with: "CodeGate Security Analysis"

If you see a string that begins with REDACTED word, DO NOT MODIFY THAT STRING while responding back.
```

yrobla: i can remove and try different workaround for Kodu. It's working partially anyway

aponcedeleonch (dismissed):
Nice!

rdimitrov (approved):

Issue #884, opened 2025-02-03
[Task]: Catch filenames from prompts


aponcedeleonch:
### Description

We have CodeSnippetExtractor which we can re-use. We need to test and validate

Issue #862, opened 2025-01-31
Support tools role in messages in the pipeline


jhrozek:
The open interpreter support in PR #820 added handling of `role: tool` messages scoped to open interpreter. This was the right thing to scope them to the tool as we wanted to get the tool support in, but we should support that message role in general in our pipeline.

more todo items would probably come out when we do the work but for now I think we should do:
 - when messages with that role are processed (e.g. `get_last_user_message_block`) do we take the `tools` role into account
 - make sure we don't touch those in pipeline to avoid breaking the tools

We can add a bunch of tests just against the openai/anthropic api to make sure we follow their APIs, no need to tie this to a specific client.

yrobla:
It would also be worth investigating how this role is being used in other tools (if that's used at all) and how to better support it

aponcedeleonch:
We should investigate which clients besides Open Interpreter are using tools to extract code snippets from files. After #945 we are able to detect when Open Interpreter does it and extract the detected code snippets. Potentially we can build something generic for detecting snippets and not Open Interpreter specific.

[See original discussion in PR](https://github.com/stacklok/codegate/pull/945#discussion_r1944293646)

jhrozek:
This will be fixed for free with the pipeline refactor -> in progress

lukehinds:
Contingent upon LiteLLM refactor

lukehinds:
closing as delivered (correct me if wrong @jhrozek )

PR #945, opened 2025-02-05
Extended the CodeSnippetExtractor functionality for muxing

aponcedeleonch:
Closes: #884 

We used to have a generic CodeSnippetExtractor to detect code snippets and the corresponding filepaths. This PR extends it's functionality to be able to detect snippets coming from different clients. After this PR the supported and tested clients are:
- Continue
- Cline
- Aider
- Open Interpreter

The main reason behind this change is to be able to catch the filepaths coming through CodeGate, detect it's destination and mux according to them. The PR also adds the necessary in the router used for muxing to detect the filepaths from the request messages.

Since CodeSnippetExtractor is no longer a pipeline step it was moved out of the folder `pipeline` to it's own folder


src/codegate/extract_snippets/body_extractor.py
```diff
@@ -0,0 +1,104 @@
...
+    DefaultCodeSnippetExtractor,
+    OpenInterpreterCodeSnippetExtractor,
+)
+
+
+class BodyCodeSnippetExtractor(ABC):
+
+    def __init__(self):
+        # Initialize the extractor in parent class. The child classes will set the extractor.
+        self._snippet_extractor: CodeSnippetExtractor = None
```
jhrozek, line 17: Nit, to satisfy pydantic we'd probably want something like:
```
 self._snippet_extractor: CodeSnippetExtractor | None = None
```
or:
```
Union[CodeSnippetExtractor, None]
```

and then check for the extractor to be set in `_extract_from_user_messages`.

src/codegate/extract_snippets/body_extractor.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _extract_from_user_messages(self, data: dict) -> set[str]:
+        copied_data = data.copy()
```
jhrozek, line 20: Do we need the copy here? Please double check this with me (I am really not sure) but the code snippets are retrieved using `match.group()` eventually.

The reason I am nitpicking about this is - when we ditch the `ChatCompletionRequest` through the work that @blkt is doing, we'll have methods to retrieve the messages from the request instead of a dict.

src/codegate/extract_snippets/body_extractor.py
```diff
@@ -0,0 +1,104 @@
...
+        return self._extract_from_user_messages(data)
+
+
+class OpenInterpreterBodySnippetExtractor(BodyCodeSnippetExtractor):
+
+    def __init__(self):
+        self._snippet_extractor = OpenInterpreterCodeSnippetExtractor()
+
+    def _is_msg_tool_call(self, msg: dict) -> bool:
+        return msg.get("role", "") == "assistant" and msg.get("tool_calls", [])
```
jhrozek, line 68: Let's add a comment to https://github.com/stacklok/codegate/issues/862 to follow up on this call - I think there's nothing OpenInterpreter specific about `tools` and we might want to remove this client special casing in favor of suporting tools in general. But we shouldn't block your work over it.

src/codegate/extract_snippets/message_extractor.py
```diff
@@ -0,0 +1,345 @@
...
+from pathlib import Path
+from typing import Dict, List, Optional, Self
+
+import structlog
+from pydantic import BaseModel, field_validator, model_validator
+from pygments.lexers import guess_lexer
+
+logger = structlog.get_logger("codegate")
+
+CODE_BLOCK_PATTERN = re.compile(
```
jhrozek, line 12: Could we move these constants into the classes?

src/codegate/extract_snippets/message_extractor.py
```diff
@@ -0,0 +1,345 @@
...
+                    unique_snippets[filename] = snippet
+
+        return unique_snippets
+
+
+class DefaultCodeSnippetExtractor(CodeSnippetExtractor):
+
+    @property
+    def codeblock_pattern(self) -> re.Pattern:
+        return [CODE_BLOCK_PATTERN]
```
jhrozek, line 281: I'm confused about us returning the list here, isn't re.Pattern just what re.compile returns?

src/codegate/extract_snippets/message_extractor.py
jhrozek: Oh do we keep them outside because of `re.compile()`?

src/codegate/pipeline/codegate_context_retriever/codegate.py
```diff
@@ -25,6 +25,9 @@ class CodegateContextRetriever(PipelineStep):
     the word "codegate" in the user message.
     """
 
+    def __init__(self):
+        self.extractor = DefaultCodeSnippetExtractor()
```
jhrozek, line 29: Should this initialze the extractor from the factory? (The per-client extractor) ?

src/codegate/pipeline/comment/output.py
```diff
@@ -17,6 +20,9 @@
 class CodeCommentStep(OutputPipelineStep):
     """Pipeline step that adds comments after code blocks"""
 
+    def __init__(self):
+        self.extractor = DefaultCodeSnippetExtractor()
```
jhrozek, line 24: I guess we don't need the per-client extractor here because all outputs are just the same format?

jhrozek (commented):
thank you, nice work, some comments and questions inline

src/codegate/extract_snippets/body_extractor.py
jhrozek: sorry that was unclear, what I mean to say above was - I think match.group already returns a copy

src/codegate/extract_snippets/message_extractor.py
aponcedeleonch: We keep them outside because of `re.compile()`. Actually, before this question I never knew the actual benefit from it. I looked for it and found this [SO post](https://stackoverflow.com/questions/452104/is-it-worth-using-pythons-re-compile). Basically the difference of compiling vs not is negligible but since we strive for performance I would keep it.

src/codegate/extract_snippets/message_extractor.py
aponcedeleonch: The only reason for returning a list is Aider. Aider can send code snippets in 2 different formats. One when it sends basically the contents of all file summarized (`AIDER_SUMMARIES_CONTENT_PATTERN`), and another one when it sends the actual content of the current file (`AIDER_FILE_CONTENT_PATTERN`).

In order to be able to match both we return them as a list in `AiderCodeSnippetExtractor`. To not break the pattern all of the extractor return a list.

src/codegate/pipeline/codegate_context_retriever/codegate.py
aponcedeleonch: Yes, it should. Will make the change. It is a an overlook on my part, sorry. I changed this at the beginning when I refactor the snippet extractor to be a class.

src/codegate/pipeline/comment/output.py
aponcedeleonch: It was an overlook on my side at the beginning. After thinking a little bit more about it I'm actually not sure, I didn't analyze the output chunks as closely as the input ones. Since before we only had a single extractor and it had the same functionality as `DefaultCodeSnippetExtractor` I would leave it as it is to not assume something wrong and break something. If we find out it doesn't work we can replace later. WDYT?

src/codegate/extract_snippets/body_extractor.py
aponcedeleonch: Nice idea, otherwise we might forget. Added the comment

yrobla (approved):

src/codegate/pipeline/comment/output.py
jhrozek: Yes, I think in output it makes sense. I haven't seen LLMs producing output in different formats and if they do, it won't be client-based.

jhrozek (approved):

PR #953, opened 2025-02-06
feat: aggregate alerts per conversation for dashboard

alex-mcgovern:
instead of returning many alerts per conversation, this wires up the `/messages` endpoint to return all alerts associated with that conversation


src/codegate/api/v1_models.py
```diff
@@ -147,6 +147,32 @@ def add_model_token_usage(self, model_token_usage: TokenUsageByModel) -> None:
         self.token_usage += model_token_usage.token_usage
 
 
+class Alert(pydantic.BaseModel):
```
aponcedeleonch, line 150: Do we want a new model to have better typing? I would modify the existing one

alex-mcgovern: I didn't want to break any existing code at this stage. But I think we could probably remove the `/alerts` endpoint and the `AlertConversation` model after we go with this approach, which would leave this as the only representation of an alert at the API layer — wdyt?

aponcedeleonch: I was thinking more like having a single model representation for both the db and the API layer. Both have the same fields. But yes, not having a single model is no reason to block the PR and once we test we can remove the redundant models.

aponcedeleonch (approved):

Issue #952, opened 2025-02-06
[Task]: Add tests for redacting secrets


rdimitrov:
### Description

Add tests for all providers for redacting secrets

PR #954, opened 2025-02-06
Add integration tests for redacting secrets from chat

rdimitrov:
The following PR adds tests for redacting secrets to all providers from chat.

Fixes: #952


jhrozek:
> I think I should comment the vllm redacting secret test or use our hosted vllm server otherwise the integration tests will take around 40minutes to complete and that's not practical.

Please do, the redaction shouldn't be depending on the provider anyway

jhrozek (approved):

Issue #885, opened 2025-02-03
[Task]: Add rules to filter by file extension or name


aponcedeleonch:
### Description

We have a single "catch all" rule. We need to start adding more rules to muxing starting with filtering by file extension and file name

Issue #944, opened 2025-02-05
[Task]: Add rules to match by type of request, i.e. chat or FIM


aponcedeleonch:
### Description

This is a low hanging fruit that can be implemented with our initial set of rules

JAORMX:
This is done

PR #968, opened 2025-02-06
Add mux rules for match according to file and request type

aponcedeleonch:
Closes: #885 and #944

This PR introduces the necessary changes for:
1. Muxing to match on a specific filename (`main.py`) or file type (`.py`)
2. Muxing on a specific request type, i.e. chat or FIM

The matching is tightly coupled with the priority of the muxing rules. A match will occur on the first possible rule.

Example:
1. Muxing rule 1 -> match all `.py` files -> go to `chat-gpt`
2. Muxing rule 2 -> match all `.md` files -> go to `ollama`

If in a request we recieve a `README.md` and a `main.py` file the request will be routed to `chat-gpt` since it's the match with the highest priority in the list.

To introduce above changes there were some minor changes:
1. Separate FIM detection into its own class. Before it used to be part of `BaseProvider`
3. Detect if a request is FIM or not before calling `process_request` method.

Both of them were necessary in order to re-use the logic to create the matcher for request type


peppescg (dismissed):
amazing 🚀  

Correct me if I am wrong, this should backward compatible with the current production API right? so if we merge it, without adding the new related UI part, it should works as well.

aponcedeleonch:
@peppescg yes, that's correct :)

JAORMX (approved):

PR #975, opened 2025-02-07
Test Pytorch install

aponcedeleonch:


aponcedeleonch:
This was just a test of the CI. Closing

aponcedeleonch:
Closing again since we're able to merge #931. The solution was to not use `pytorch` and use `onnx` for inference instead.

Issue #969, opened 2025-02-06
Error loading prompt history and opening events


danbarr:
### Describe the issue

This is happening only when I run from `main` (I built the container with `make all && make image-build` and ran that version). Reverting to the `v0.1.16` release fixes it.

When I open the dashboard, the prompt history fails to load, and I can't open events (goes to a blank screen). In the container logs I see the following error, repeated several times per attempt:

```
2025-02-06T23:33:20.397389Z [error    ] Error while getting messages   lineno=424 module=v1 pathname=/app/src/codegate/api/v1.py
  + Exception Group Traceback (most recent call last):
  |   File "/app/src/codegate/api/v1.py", line 419, in get_workspace_messages
  |     conversations, _ = await v1_processing.parse_messages_in_conversations(
  |                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/app/src/codegate/api/v1_processing.py", line 436, in parse_messages_in_conversations
  |     partial_question_answers = await _process_prompt_output_to_partial_qa(prompts_outputs)
  |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/app/src/codegate/api/v1_processing.py", line 425, in _process_prompt_output_to_partial_qa
  |     async with asyncio.TaskGroup() as tg:
  |                ^^^^^^^^^^^^^^^^^^^
  |   File "/usr/local/lib/python3.12/asyncio/taskgroups.py", line 71, in __aexit__
  |     return await self._aexit(et, exc)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "/usr/local/lib/python3.12/asyncio/taskgroups.py", line 164, in _aexit
  |     raise BaseExceptionGroup(
  | ExceptionGroup: unhandled errors in a TaskGroup (57 sub-exceptions)
```

I'll include the full traceback message below since it's very long.

### Steps to Reproduce

With existing prompt data in database (I can provide a copy of mine if needed), start codegate from main.

### Operating System

MacOS (Arm)

### IDE and Version

NA

### Extension and Version

NA

### Provider

Anthropic

### Model

NA

### Codegate version

v0.1.16-91b73bf-dev

### Logs

[messages-error.txt](https://github.com/user-attachments/files/18698977/messages-error.txt)

PR #976, opened 2025-02-07
Fixe on Pydantic model for alerts

aponcedeleonch:
Closes: #969

Some small fixes all related with fixing alerts both when adding from secrets and when retrieving them through the endpoint `http://localhost:8989/api/v1/workspaces/openai/messages`

### DIFF
--- a/src/codegate/db/models.py
+++ b/src/codegate/db/models.py
@@ -5,14 +5,19 @@ from typing import Annotated, Any, Dict, List, Optional
 from pydantic import BaseModel, StringConstraints
 
 
+class AlertSeverity(str, Enum):
+    INFO = "info"
+    CRITICAL = "critical"
+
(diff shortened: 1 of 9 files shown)


src/codegate/db/models.py
```diff
@@ -5,14 +5,19 @@
...
-    trigger_type: Any
-    trigger_category: Optional[Any]
-    timestamp: Any
+    id: str
+    prompt_id: str
+    code_snippet: Optional[str]
+    trigger_string: Optional[str]
+    trigger_type: str
+    trigger_category: AlertSeverity
+    timestamp: datetime.datetime
```
aponcedeleonch, line 20: Be more explicit in our typing with alerts

peppescg (approved):

Issue #981, opened 2025-02-07
OpenRouter+DeepSeek doesn't extract codegate CLI tasks


jhrozek:
Other providers work well, but with OpenRouter and Continue, `codegate version` just falls through.

PR #982, opened 2025-02-07
Fix codegate version for continue with DeepSeek

jhrozek:
`continue version` didn't work with DeepSeek because for some reason
they enclose the user message in a sort of an envelope.

To make the change a little future-proof, we use the envelope to detect
Continue as client and then in the cli command try to extract from the
envelope, but fall back to the generic plain regex.

Fixes: #981

### DIFF
--- a/src/codegate/pipeline/cli/cli.py
+++ b/src/codegate/pipeline/cli/cli.py
@@ -13,6 +13,8 @@ from codegate.pipeline.base import (
 )
 from codegate.pipeline.cli.commands import CustomInstructions, Version, Workspace
 
+codegate_regex = re.compile(r"^codegate(?:\s+(.*))?", re.IGNORECASE)
+
 HELP_TEXT = """
 ## CodeGate CLI\n
 **Usage**: `codegate [-h] <command> [args]`\n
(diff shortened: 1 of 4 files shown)


src/codegate/pipeline/cli/cli.py
```diff
@@ -77,6 +79,22 @@ def _get_cli_from_open_interpreter(last_user_message_str: str) -> Optional[re.Ma
     return re.match(r"^codegate\s*(.*?)\s*$", last_user_block, re.IGNORECASE)
 
 
+def _get_cli_from_continue(last_user_message_str: str) -> Optional[re.Match[str]]:
+    """
+    Continue sends a differently formatted message to the CLI if DeepSeek is used
+    """
+    deepseek_match = re.search(
```
aponcedeleonch, line 86: I saw this happening with other model (`tinyllama:latest`). I was testing muxing so in reality I was hitting an open AI model but that was what I had in my Continue config.

Anyways just mentioning it because the var name implies is specific to deepseek and it maybe a gradual release in general for Continue.

aponcedeleonch (approved):

PR #983, opened 2025-02-07
Fix to also prepend openrouter/ when muxing

aponcedeleonch:
Moved inside a function and called parent


aponcedeleonch:
This works but I need to fix the tests

Issue #875, opened 2025-01-31
CodeGate seems to be raising more alerts than expected


wright-io:
### Describe the issue

The latest release of CodeGate seems to be raising more alerts than expected . When the copliot response says there are 2 secrets protected, 6 are showing up in the dashboard (including firing chrome alerts, if configured). 

### Steps to Reproduce

- Ask copilot with codegate to review https://github.com/stacklok/codegate-demonstration/blob/main/conf.ini (contains 3 secrets)
- Copilot responds "🛡️ CodeGate prevented 2 secrets from being leaked by redacting them"
- Dashboard contains 6 alerts (see screenshot)

<img width="1208" alt="Image" src="https://github.com/user-attachments/assets/7d24399f-f8b7-43eb-98d2-530338de4bfd" />

### Operating System

MacOS (Arm)

### IDE and Version

-

### Extension and Version

-

### Provider

OpenAI

### Model

GPT-4o

### Codegate version

v0.1.15

yrobla:
that is because actually github copilot generates internal messages , splitting the request in two. I added some logic to group those requests in one, i will do the same for alerts

PR #989, opened 2025-02-07
feat: remove duplicated alerts

yrobla:
Sometimes the different client tools generate multiple requests when the user requests a task. This generates what it looks as a duplicate alert, but is not really a total duplicate as it belongs to different request.
But for the user does not provide so much value to have it, so proceed with deduplicating those alerts, based on the code snippet and details of the alert

Also remove dogecoin regex as it's giving false positives

Closes: #875

### DIFF
--- a/signatures.yaml
+++ b/signatures.yaml
@@ -281,7 +281,6 @@
     - Bitcoin SegWit: \b(bc1)[a-zA-HJ-NP-Z0-9]{39,59}\b
     - Ethereum: \b0x[a-fA-F0-9]{40}\b
     - Litecoin: \b(L|M)[a-km-zA-HJ-NP-Z1-9]{26,33}\b
-    - Dogecoin: \b(D|A)[a-km-zA-HJ-NP-Z1-9]{25,34}\b
     - Ripple: \br[rK][a-zA-Z0-9]{25,35}\b
     - Monero: \b4[0-9AB][1-9A-HJ-NP-Za-km-z]{93}\b
     - Tron: \bT[a-zA-HJ-NP-Z0-9]{33}\b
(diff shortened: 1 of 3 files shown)


src/codegate/api/v1.py
```diff
@@ -390,6 +390,7 @@ async def get_workspace_alerts(workspace_name: str) -> List[Optional[v1_models.A
 
     try:
         alerts = await dbreader.get_alerts_by_workspace(ws.id, AlertSeverity.CRITICAL.value)
+        alerts = v1_processing.remove_duplicate_alerts(alerts)
```
JAORMX, line 393: Should this be part of `get_alerts_by_workspace`? Having to do an extra call seems misleading and easy to miss.

JAORMX (changes requested):

signatures.yaml
```diff
@@ -281,7 +281,6 @@
     - Bitcoin SegWit: \b(bc1)[a-zA-HJ-NP-Z0-9]{39,59}\b
     - Ethereum: \b0x[a-fA-F0-9]{40}\b
     - Litecoin: \b(L|M)[a-km-zA-HJ-NP-Z1-9]{26,33}\b
-    - Dogecoin: \b(D|A)[a-km-zA-HJ-NP-Z1-9]{25,34}\b
```
JAORMX, line 284: Was this intended? Seems unrelated.

JAORMX (changes requested):

src/codegate/api/v1.py
aponcedeleonch: Yes, nice idea. Moved it there

signatures.yaml
aponcedeleonch: Part of another issue. Yes, it's intended, will separate in a different PR

JAORMX (approved):

PR #1001, opened 2025-02-10
Generate valid chunked response when returning short circuit answer

dmjb:
Prior to this change, this code would return a malformed HTTP response with a chunk size, then HTTP header, then body, then the final chunk size, final chunk then zero.

This changes the code to return the header before returning the code. Furthermore, it does not count the size of the header in the first chunk size.

This has been validated as a fix in the scenario where I encountered it, plus in contexts where this flow worked previously.


jhrozek (commented):
I think this patch makes sense and all my testing in vscode works fine.

jhrozek (approved):

PR #1002, opened 2025-02-10
Pass down base URL and API key to completion handler

JAORMX:
In order to make Ollama work with API Keys, I needed to change the
completion handler to take a base URL and also leverage a given API key
(if available).

Co-Authored-by: Alejandro Ponce de Leon <aponcedeleonch@stacklok.com>

### DIFF
--- a/src/codegate/providers/ollama/completion_handler.py
+++ b/src/codegate/providers/ollama/completion_handler.py
@@ -82,17 +82,26 @@ async def ollama_stream_generator(  # noqa: C901
 
 class OllamaShim(BaseCompletionHandler):
 
-    def __init__(self, base_url):
-        self.client = AsyncClient(host=base_url, timeout=300)
-
     async def execute_completion(
         self,
         request: ChatCompletionRequest,
+        base_url: Optional[str],
         api_key: Optional[str],
         stream: bool = False,
         is_fim_request: bool = False,
     ) -> Union[ChatResponse, GenerateResponse]:
         """Stream response directly from Ollama API."""
+
+        if not base_url:
+            raise ValueError("base_url is required for Ollama")
+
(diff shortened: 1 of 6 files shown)


src/codegate/providers/ollama/completion_handler.py
```diff
@@ -82,17 +82,26 @@ async def ollama_stream_generator(  # noqa: C901
...
         request: ChatCompletionRequest,
+        base_url: Optional[str],
         api_key: Optional[str],
         stream: bool = False,
         is_fim_request: bool = False,
     ) -> Union[ChatResponse, GenerateResponse]:
         """Stream response directly from Ollama API."""
+
+        if not base_url:
+            raise ValueError("base_url is required for Ollama")
```
jhrozek, line 96: could we default to `localhost:11434` ?

src/codegate/providers/ollama/completion_handler.py
```diff
@@ -82,17 +82,26 @@ async def ollama_stream_generator(  # noqa: C901
...
+
+        # TODO: Add CodeGate user agent.
+        headers = dict()
+        if api_key:
+            headers["Authorization"] = f"Bearer {api_key}"
```
jhrozek, line 101: Does the bearer token get forwarded or does it have some other meaning later on?

JAORMX: It gets set so we can forward it

src/codegate/providers/ollama/completion_handler.py
JAORMX: We should already get this default frm the caller.

jhrozek (approved):
I kicked the tests, that looked like a transient error

PR #1014, opened 2025-02-11
Enable provider model updating when updating provider itself

JAORMX:
This now calls the model updating logic when updating the provider
itself. Thus allowing us to have a way to update the model list.


aponcedeleonch (changes requested):
We still need to update the models of a provider at server startup, i.e. in function `initialize_provider_endpoints`. Right now, if the provider already exists it will not try to get its models.

aponcedeleonch (approved):
Approved. The remaining functionality will be done in a separate PR

Issue #1005, opened 2025-02-10
Muxing issues with FIM in Continue


danbarr:
### Describe the issue

Possibly related to #980, but seeing problems with muxing and FIM in Continue beyond just OpenRouter.

With OpenRouter, seeing the same problem via muxing as we see in #980, CodeGate appears to be processing the FIM event but nothing shows up from the extension.

With Ollama (via mux), no error but also no output, and the logs are suspiciously thin, the prompt isn't logged at all and doesn't show up in the dashboard history.
```
2025-02-10T20:31:55.027709Z [info     ] No particilar client detected, using generic client lineno=244 module=detector pathname=/app/src/codegate/clients/detector.py
2025-02-10T20:31:55.028218Z [info     ] Catch all rule matched         lineno=90 module=rulematcher pathname=/app/src/codegate/muxing/rulematcher.py
2025-02-10T20:31:55.028354Z [info     ] Muxing request routed to destination provider lineno=91 model=qwen2.5-coder:7b module=router pathname=/app/src/codegate/muxing/router.py provider_name=Ollama provider_type=ollama
2025-02-10T20:31:55.028802Z [info     ] FIM pipeline selected for execution. lineno=154 module=base pathname=/app/src/codegate/providers/base.py
2025-02-10T20:31:55.037993Z [info     ] Total secrets redacted since last assistant message: 0 lineno=361 module=secrets pathname=/app/src/codegate/pipeline/secrets/secrets.py
2025-02-10T20:31:55.075064Z [info     ] Total PII instances redacted: 0 lineno=96 module=pii pathname=/app/src/codegate/pipeline/pii/pii.py
2025-02-10T20:31:55.075440Z [info     ] FIM pipeline selected for output. lineno=109 module=base pathname=/app/src/codegate/providers/base.py
2025-02-10T20:31:55.075511Z [info     ] No output pipeline steps configured, passing through lineno=123 module=base pathname=/app/src/codegate/providers/base.py
```

With OpenAI via the mux endpoint, I get an error:
```
HTTP 400 Bad Request from http://127.0.0.1:8989/v1/mux/completions {"detail":"litellm.BadRequestError: OpenAIException - Error code: 400 - {'error': {'message': \"Invalid 'messages': empty array. Expected an array with minimum length 1, but got an empty array instead.\", 'type': 'invalid_request_error', 'param': 'messages', 'code': 'empty_array'}}"}
```

I don't have an Anthropic account to test with.

### Steps to Reproduce

[...]

PR #1015, opened 2025-02-11
Fix FIM on Continue. Have specific formatters for chat and FIM

aponcedeleonch:
Closes: #1005 

Until now we had a general formatter on the way out of muxing. This is wrong since sometimes the pipelines respond with different format for chat or FIM. Such is the case for Ollama.

This PR separates the formatters and declares them explicitly so that they're easier to adjust in the future.

### DIFF
--- a/src/codegate/muxing/adapter.py
+++ b/src/codegate/muxing/adapter.py
@@ -3,6 +3,7 @@ import json
 import uuid
 from abc import ABC, abstractmethod
 from typing import Callable, Dict, Union
+from urllib.parse import urljoin
 
 import structlog
 from fastapi.responses import JSONResponse, StreamingResponse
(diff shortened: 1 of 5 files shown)


aponcedeleonch:
Still testing. Opened a PR to check if I hadn't broken something using CI

jhrozek (dismissed):
I know this is a draft, but I think this patch makes sense and in general I think this is caused by openai python libraries being bumped in litellm. I noticed that the new versions don't support prompt anymore.

src/codegate/muxing/adapter.py
```diff
@@ -30,12 +31,13 @@ class BodyAdapter:
...
         """Get the provider formatted URL to use in base_url. Note this value comes from DB"""
+        base_endpoint = model_route.endpoint.endpoint.rstrip("/")
         if model_route.endpoint.provider_type in [
             db_models.ProviderType.openai,
             db_models.ProviderType.openrouter,
         ]:
-            return f"{model_route.endpoint.endpoint}/v1"
-        return model_route.endpoint.endpoint
+            return f"{base_endpoint}/v1"
+        return base_endpoint
```
JAORMX, line 40: nit: instead of doing raw text manipulation, should we instead be using `urlpath.join`?

src/codegate/providers/ollama/adapter.py
```diff
@@ -47,21 +46,36 @@ def __init__(self, ollama_response: AsyncIterator[ChatResponse]):
...
+        role = "assistant"
+        if is_chunk_done:
             finish_reason = "stop"
             role = None
+        return role, finish_reason
+
+    @classmethod
+    def _get_chat_id_from_timestamp(cls, timestamp_seconds: int) -> str:
+        timestamp_str = str(timestamp_seconds)
+        return timestamp_str[:9]
```
JAORMX, line 67: This merits a bit of documentation.

JAORMX (approved):
minor comments, feel free to address them in subsequent PRs

src/codegate/muxing/adapter.py
aponcedeleonch: Yes, nice idea. Fixing

src/codegate/providers/ollama/adapter.py
aponcedeleonch: Agreed, adding

JAORMX (approved):

PR #1023, opened 2025-02-12
Relax OpenAI models to align with upstream spec

jhrozek:
We were too strict with our models. Let's relax them, see e.g. https://github.com/openai/openai-python/blob/main/src/openai/types/chat/chat_completion_chunk.py


jhrozek:
@blkt 👀

blkt (approved):

PR #1026, opened 2025-02-12
Add set_text to Anthropic UserMessage to enable raw string messages

jhrozek:
In case the message is just a text and not a MessageContent, we need to set self.content. Otherwise, the method will be called on MessageContent which is already implemented.


jhrozek:
@blkt 👀

blkt (approved):

PR #1031, opened 2025-02-12
Methods for openai reply types

jhrozek:
This allows using the openai types without normalizers.


jhrozek:
@blkt crap, I'm sorry, I clicked the merge button by accident. Can you review nonetheless, I'll be happy to fix whatever is needed.

Issue #841, opened 2025-01-30
[Task]: Add integration test for comparing the result of calling the API directly vs through CodeGate


rdimitrov:
### Description

Update the integration tests so one can ensure that CodeGate is not enriching the context and thus affecting the request and response when it shouldn't.

PR #1032, opened 2025-02-13
Add integration checks for comparing the result of calling the model API directly vs via CodeGate

rdimitrov:
This introduces the following format:

```yaml
    codegate_enrichment:
      provider_url: http://127.0.0.1:11434/api/chat
      expect_difference: false
```

Fixes: #841

### DIFF
--- a/tests/integration/integration_tests.py
+++ b/tests/integration/integration_tests.py
@@ -9,11 +9,24 @@ from typing import Any, Dict, Optional, Tuple
 import requests
 import structlog
 import yaml
-from checks import CheckLoader
+from checks import CheckLoader, CodeGateEnrichment
 from dotenv import find_dotenv, load_dotenv
 from requesters import RequesterFactory
 
 logger = structlog.get_logger("codegate")
+
+
+# call_directly is a function to call the model directly bypassing codegate
+def call_directly(url: str, headers: dict, data: dict) -> Optional[requests.Response]:
+    try:
+        headers["Content-Type"] = "application/json"
+        stream = data.get("stream", False)
+        response = requests.post(url, headers=headers, json=data, stream=stream)
+        response.raise_for_status()
+        return response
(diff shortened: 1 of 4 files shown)


tests/integration/integration_tests.py
```diff
@@ -9,13 +9,26 @@
...
+def call_directly(url: str, headers: dict, data: dict) -> Optional[requests.Response]:
+    try:
+        headers["Content-Type"] = "application/json"
+        stream = data.get("stream", False)
+        response = requests.post(url, headers=headers, json=data, stream=stream)
+        response.raise_for_status()
+        return response
+    except Exception as e:
+        logger.error(f"Error making direct request to {url}: {str(e)}")
+        return None
```
aponcedeleonch, line 29: I think this code is almost identical as `call_codegate`. Could we reuse/rename that one? Anyways it's not a big deal.

aponcedeleonch (dismissed):
Just one comment. But in general the code looks good. Out of curiosity, are we not going to test also with OpenAI, Anthropic and Openrouter?

tests/integration/integration_tests.py
rdimitrov: ah, good point, I'll fix it 👍

rdimitrov:
> Just one comment. But in general the code looks good. Out of curiosity, are we not going to test also with OpenAI, Anthropic and Openrouter?
My understanding is that it is enough to ensure this for one provider and so it doesn't make sense to repeat it for others (and consume API tokens). Is this correct assumption or we should extend these tests for the others too?

aponcedeleonch:
> My understanding is that it is enough to ensure this for one provider and so it doesn't make sense to repeat it for others (and consume API tokens). Is this correct assumption or we should extend these tests for the others too?

Yes, that's right

aponcedeleonch (approved):

Issue #1039, opened 2025-02-13
[Task]: Suspicious commands needs its own pipeline


lukehinds:
### Description

Suspicious commands is not implemented into the pipeline, its instead hooking into the main pipeline output:

https://github.com/stacklok/codegate/blob/main/src/codegate/pipeline/comment/output.py#L54

It should instead be a pipeline object in the same way as PII, secrets etc.

therealnb:
The code was generally in the right place and generally working. Secrets and PII were being found, but this was real.

Shell languages were not being recognised, this is alleviated here https://github.com/stacklok/codegate/pull/1043/files#diff-f7211151eec890242ec818bcdda3b3a175dda1c686533e679771037692952d5fL141

In copilot edits, the deltas were being passed as snippets. We have alleviated that here https://github.com/stacklok/codegate/pull/1043/files#diff-0658dbd5db53af1b59ad53ac8ce9dc0ba40042abea62cce7f6b04eebf3fa4cffR63
So that suspicious commands will not be applied to the main programming languages.

Both of these will have corner cases, but [these](https://github.com/stacklok/codegate/pull/1043) fixes offer a low impact set of changes.

therealnb:
Also reported this https://github.com/stacklok/codegate/issues/1044
CC @jhrozek

therealnb:
Note that we had to disable this code https://github.com/stacklok/codegate/pull/1073

It is now unclear where this code should go.

PR #1043, opened 2025-02-13
Handle shell languages slightly better

therealnb:
Closes https://github.com/stacklok/codegate/issues/1039
See the ticket for details


jhrozek (approved):

jhrozek:
Approved Nigel showed me that the change improves the detection on the phone

PR #1063, opened 2025-02-14
Only handle user messages with text, not None

jhrozek:
Tool calls have no content, we should not handle them. This bug triggered issues with the agentic copilot.


jhrozek:
@blkt

JAORMX (approved):

blkt (approved):

PR #1064, opened 2025-02-14
Modify to use GPU (if available) for embedding

ptelang:


JAORMX (approved):

ptelang:
> Could I get a little more context Pankaj, is this meaning its using metal or cuda (or both)?

It will use either metal or cuda, whichever is available. The actual implementation of using the GPU is in llama.cpp. GPU was not getting used since codegate was not passing a parameter to the llama.cpp call. This PR fixes the issue by passing that parameter.

JAORMX:
Good find @ptelang !

PR #1074, opened 2025-02-17
Secrets simple

jhrozek:
- fixes a typo
- removes a condition that seems to not be needed

@blkt I'm less sure about the second one, do you agree with that change?


blkt (approved):
I think we can merge it all, I have tests for the `secrets` step running locally, we can fix it later if they don't pass.

PR #1080, opened 2025-02-17
Dump raw JSON messages

jhrozek:
This amends the helper that dumps the requests to dump the JSON as represented by the models instead of a printable representation of the models.


jhrozek:
@blkt 👀

blkt (approved):

PR #1081, opened 2025-02-17
Amend the context replacer for the new data structures

jhrozek:
I'll point out what's going on inline

### DIFF
--- a/src/codegate/pipeline/codegate_context_retriever/codegate.py
+++ b/src/codegate/pipeline/codegate_context_retriever/codegate.py
@@ -26,6 +26,21 @@ class CodegateContextRetriever(PipelineStep):
     the word "codegate" in the user message.
     """
 
+    def __init__(
+            self,
+            storage_engine: StorageEngine | None = None,
+            package_extractor: PackageExtractor | None = None,
+        ):
+        """
+        Initialize the CodegateContextRetriever with optional dependencies.
+
+        Args:
+            storage_engine: Optional StorageEngine instance for package searching
+            package_extractor: Optional PackageExtractor class for package extraction
+        """
+        self.storage_engine = storage_engine or StorageEngine()
+        self.package_extractor = package_extractor or PackageExtractor
+
     @property
     def name(self) -> str:
         """
@@ -67,9 +82,6 @@ class CodegateContextRetriever(PipelineStep):
             return PipelineResult(request=request)
         user_message, last_user_idx = last_message
 
-        # Create storage engine object
-        storage_engine = StorageEngine()
-
         # Extract any code snippets
         extractor = MessageCodeExtractorFactory.create_snippet_extractor(context.client)
         snippets = extractor.extract_snippets(user_message)
@@ -81,7 +93,7 @@ class CodegateContextRetriever(PipelineStep):
             snippet_packages = []
             for snippet in snippets:
                 snippet_packages.extend(
-                    PackageExtractor.extract_packages(snippet.code, snippet.language)  # type: ignore
+                    self.package_extractor.extract_packages(snippet.code, snippet.language)  # type: ignore
                 )
 
             logger.info(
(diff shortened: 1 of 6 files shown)


src/codegate/types/anthropic/_request_models.py
```diff
@@ -110,7 +110,7 @@ def get_text(self) -> Iterable[str]:
 
     def set_text(self, text) -> None:
         if isinstance(self.content, str):
-            self.content = txt
+            self.content = text
```
jhrozek, line 113: this should probably be its own patch, but eh. it's just a typo fix.

src/codegate/types/openai/_request_models.py
```diff
@@ -339,9 +339,9 @@ def last_user_message(self) -> tuple[Message, int] | None:
 
     def last_user_block(self) -> Iterable[tuple[Message, int]]:
         for idx, msg in enumerate(reversed(self.messages)):
-            if isinstance(msg, UserMessage):
+            if isinstance(msg, (UserMessage, ToolMessage)):
```
jhrozek, line 342: This is to be able to extract bad packages from files read by the tools, see the unit test for an example I grabbed from OpenInterpreter session. I totally forgot that tool call results can contain bad packages...

src/codegate/pipeline/codegate_context_retriever/codegate.py
```diff
@@ -130,42 +130,34 @@ async def process(  # noqa: C901
...
+                            f"<task>Context: {context_str}"
+                            + f"Query: {task_content.strip()}</task>"
+                        )
+
+                        # Combine updated task content with the rest of the message
+                        context_msg = updated_task_content + rest_of_message
+                else:
+                    context_msg = f"Context: {context_str} \n\n Query: {message_str}"
+                content = next(message.get_content())
+                content.set_text(context_msg)
```
jhrozek, line 178: With the set_text call we can just drop the open-interpreter special-case. I tested it by asking OI to evaluate dependencies in requirements.txt.

src/codegate/pipeline/codegate_context_retriever/codegate.py
```diff
@@ -131,11 +131,13 @@ async def process(  # noqa: C901
...
-                    txt
-                    for content in message.get_content()
-                    for txt in content.get_text()
-                ])
+                message_str = ""
+                for content in message.get_content():
+                    txt = content.get_text()
+                    if not txt:
+                        logger.debug(f"content has no text: {content}")
+                        continue
```
jhrozek, line 155: I found a traceback while asking Continue+Anthropic to evaluate dependencies in requirements.txt. There is also a unit test with exactly those contents.

tests/pipeline/codegate_context_retriever/test_codegate.py
```diff
@@ -0,0 +1,319 @@
...
+                    "name": "archived-package-1",
+                    "type": "npm",
+                    "status": "archived",
+                    "description": "This package is archived-1",
+                },
+            },
+        ]
+        mock_storage_engine.search = AsyncMock(return_value=bad_packages)
+
+        request = AnthropicChatCompletionRequest(
```
jhrozek, line 243: this is a model based on one test that I used to provoke a traceback

tests/pipeline/test_messages_block.py
```diff
@@ -123,63 +123,62 @@
...
...e, lmc=None, pil_image=None) # Gets OCR of image.\ncomputer.vision.query(query='Describe this image. Also tell me what text is in the image, if any.', base_64=None, path=None, lmc=None, pil_image=None) # Uses Moondream to ask query of the image (which can be a base64, path, or lmc message)\ncomputer.skills.import_skills() # \ncomputer.skills.list() # \ncomputer.skills.run(skill) # \ncomputer.skills.search(query) # This just lists all for now.\ncomputer.docs.search(query, module=None, paths=None) # \ncomputer.ai.chat(text, base64=None) # \ncomputer.ai.query(text, query, custom_reduce_query=None) # \ncomputer.ai.summarize(text) # \ncomputer.files.edit(path, original_text, replacement_text) # Edits a file on the filesystem, replacing the original text with the replacement text.\ncomputer.files.search() # Search the filesystem for the given query.\n```\n\nDo not import the computer module, or any of its sub-modules. They are already imported.\n\nThe files in the context contain sensitive information that has been redacted. Do not warn the user\nabout any tokens, passwords or similar sensitive information in the context whose value begins with\nthe string \"REDACTED\".\n",  # noqa: E501
```
jhrozek, line 133: The test now works insofar as the test case returns only the tool output, because the message before it is an assistant message. I still need to crack aider, it's tricky because they inject assistant messages.

blkt (approved):

PR #1088, opened 2025-02-18
Add support for runtime override of dashboard url.

blkt:
I opened this PR to discuss whether we want to do it and, if so, to keep it as simple as possible.


JAORMX (approved):
Seems like a reasonable first step and it would unblock me

davolokh:
agree, looks like a pragmatic solution

PR #1092, opened 2025-02-18
Remove now-unused code from the code-comment pipeline step

jhrozek:
I tested that the piece works, really the only changes are removal of code that is no longer used.


jhrozek:
@blkt 👀

blkt (approved):

JAORMX (approved):

Issue #977, opened 2025-02-07
CodeGate doesn't include reasoning for OpenRouter thinking models


jhrozek:
It's possible to follow the reasoning of the thinking models like DeepSeek through OpenRouter, see https://openrouter.ai/announcements/reasoning-tokens-for-thinking-models

However, that doesn't work through codegate it seems.

jhrozek:
Looks like we need to make sure that the `"include_reasoning": true` flag is included in the request. Hopefully, litellm will not mangle that again.

jhrozek:
so on our side, the fix is simple, but won't work until https://github.com/BerriAI/litellm/issues/8193 is fixed.

jhrozek:
parking

PR #1096, opened 2025-02-19
Pass through reasoning input flag and output chunks

jhrozek:
Some OpenRouter models allow to include the reasoning output if told to do so with a special input flag. This seems to be specific to OpenRouter.

A simple test is:
```
curl http://localhost:8989/openrouter/completions \
 -H "Authorization: Bearer $OPENROUTER_KEY" \
 -H "Content-Type: application/json" \
 -d '{
   "model": "deepseek/deepseek-r1",
   "messages": [
     {"role": "user", "content": "What is heavier, a kilogram of iron or a kilogram of feathers?"}
   ],
   "include_reasoning": true,
   "stream": true
 }'
```

Fixes: #977

### DIFF
--- a/src/codegate/types/openai/_request_models.py
+++ b/src/codegate/types/openai/_request_models.py
@@ -318,6 +318,7 @@ class ChatCompletionRequest(pydantic.BaseModel):
     user: str | None = None
     function_call: str | FunctionChoice | None = "auto" # deprecated
     functions: List[LegacyFunctionDef] | None = None # deprecated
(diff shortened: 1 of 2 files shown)


jhrozek:
I should note that I didn't find it easy to trigger the reasoning output with Cline lately. I remember seeing it with vanilla openrouter (not through codegate), but I have not during this testing. I see the chunks during curl testing and they seem the same as when directly going to openrouter. Maybe it's just the queries I've been sending to Cline?

blkt (approved):

PR #1097, opened 2025-02-19
Fix FIM for OpenRouter

aponcedeleonch:
FIM was not working with OpenRouter in Continue. The reason was that we get FIM requests in `/completions`. LiteLLM when using `acompletion` is forcing `/chat/completions` and the return format `{..., "choices":[{"delta":{"content":"some text"}}]}`. However, Continue was expecting the format: `{..., "choices":[{"text":"some text"}]}` becuase of the endpoint it called.

With this PR we force the return format to be the latter using `atext_completion` from LiteLLM

### DIFF
--- a/src/codegate/providers/openai/provider.py
+++ b/src/codegate/providers/openai/provider.py
@@ -18,8 +18,9 @@ class OpenAIProvider(BaseProvider):
     def __init__(
         self,
         pipeline_factory: PipelineFactory,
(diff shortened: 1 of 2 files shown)


jhrozek (dismissed):
oh, litellm. The gift that keeps on giving.

jhrozek (approved):

PR #1099, opened 2025-02-19
feat: add workspaces list endpoint by provider id

peppescg:
Added an endpoint `/workspaces/:provider_id` that return the list of workspaces by a provider id. This allow the dashboard to show the impacted workspaces in case of providers deletion. The dashboard will show a list of workspaces in the confirmation dialog before a provider will be deleted, in this way it is clear which workspaces are impacted.

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -534,7 +534,7 @@ async def set_workspace_muxes(
 
 @v1.get(
     "/workspaces/{provider_id}",
-    tags=["Providers"],
+    tags=["Workspaces"],
     generate_unique_id_function=uniq_name,
 )
 async def list_workspaces_by_provider(
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -541,16 +541,9 @@ async def list_workspaces_by_provider(
     provider_id: UUID,
 ) -> List[WorkspaceWithModel]:
     """List workspaces by provider ID."""
-
     try:
-        workspaces = await wscrud.workspaces_by_provider(provider_id)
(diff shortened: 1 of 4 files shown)


src/codegate/api/v1.py
```diff
@@ -532,6 +532,29 @@ async def set_workspace_muxes(
...
+) -> List[WorkspaceWithModel]:
+    """List workspaces by provider ID."""
+
+    try:
+        workspaces = await wscrud.workspaces_by_provider(provider_id)
+
+        return [
+            WorkspaceWithModel(id=ws.id, name=ws.name, provider_model_name=ws.provider_model_name)
+            for ws in workspaces
+        ]
```
aponcedeleonch, line 551: `wscrud.workspaces_by_provider` is already returning a `List[WorkspaceWithModel]`. This list comprehension can be removed. In other words, it can be simply
```python
return await wscrud.workspaces_by_provider(provider_id)
```

src/codegate/api/v1.py
```diff
@@ -532,6 +532,29 @@ async def set_workspace_muxes(
...
+    except crud.WorkspaceDoesNotExistError:
+        raise HTTPException(status_code=404, detail="Workspaces not found")
```
aponcedeleonch, line 553: nit: The code is not raising `crud.WorkspaceDoesNotExistError` in any place. We would never fall into this `except` clause. If we want some pre-validation of the data it would be good to check if we have a provider with the specified `provider_id`. Although I think the code will work as is and if the provider doesn't exist the list will just return empty

aponcedeleonch (commented):
Couple of comments :)

aponcedeleonch (approved):

Issue #1066, opened 2025-02-15
Investigate CodeGate Deployment on Kubernetes


lukehinds:
investigate deploying CodeGate to Kubernetes cluster. This includes researching the necessary changes and configurations required to support this deployment method , such as dashboard access, network. This could be delivered as either / or docs or a helm chart etc.

JAORMX:
The main thing to consider here is the db deployment. I can help also if needed.

JAORMX:
I can pair with @ChrisJBurns on this

jhrozek:
@blkt and @peppescg already looked into a remote UI they might know some issues already

blkt:
This thread has some context that might help. https://discord.com/channels/1184987096302239844/1339150207979360307

ChrisJBurns:
We have a codegate helm chart, we have tested a deployment into one of our clusters and can confirm it works and is responding to healthchecks via port-forwarding. There are a couple of things that would need to be ironed out to support a more production ready implementation:
- Is only limited to 1 replica. Anymore and there will be clashes in the single file datastore due to the simple design that codegate was intended for. Perhaps a cloud hosted version would need to be less simple and allow for scaling, perhaps with something like https://github.com/rqlite/rqlite ?
- A more production ready solution would have to include a proper ingress solution instead of port forwarding

PR #1102, opened 2025-02-19
feat: adds codegate helm chart

ChrisJBurns:
- adds the initial version of the helm chart in relation to achieve a Kubernetes installation
- adds ci test for helm chart and helm chart publish to stacklok gchr

https://github.com/stacklok/codegate/issues/1066


deploy/charts/codegate/templates/deployment.yaml
```diff
@@ -0,0 +1,69 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "codegate.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "codegate.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ .Values.replicaCount }}
```
JAORMX, line 10: have we tried this with multiple replicas yet? This might conflict with the single-file sqlite database we have. Maybe we should limit it to 1.

ChrisJBurns: Nope not with multiple, defo should limit to `1`. Deployment with a PVC also forces the constraint of running it with `1`. Unless we use stateful sets but then the single file db isn't really a thing then

JAORMX (approved):

rdimitrov (approved):

JAORMX (approved):

PR #1103, opened 2025-02-19
Initial documentation on debugging issues with clients

aponcedeleonch:
Based on a recent experience with Coninue I think  we should start documenting how the issue was solved to help the next person trying to debug a similar issue.


rdimitrov (approved):
nice!

Issue #1067, opened 2025-02-15
Investigate shareable workspaces


lukehinds:
Investigate a simple shareable workspace configuration, where a user could export a workspace, capturing prompts , provider config and muxing rules. This could be saved as file which someone else could then import.

JAORMX:
I could look at this. Provider config is not workspace-scoped, but I'm sure we could make it work

peppescg:
I think this will impact the UI dashboard too. Cause I would expect to have a button in the dashboard and import/export the config ( keep in mind that muxing is actually available only in UI).

JAORMX:
We agreed someone else will work on this. I can tag along and help out if needed.

JAORMX:
I'll pair up with @alex-mcgovern

alex-mcgovern:
Actually picking this up now

alex-mcgovern:
I think a north star for this should be a local JSON "CodeGate config", this would allow:
- hand editing in editor-of-choice
- checking in to version control (either per-repo or a `.dotfiles` repo)
- sharing/distributing within a team

I'm imagining this could be stored in a number of places, e.g. `$HOME/.config/codegate`, in the currently open repository, etc. but that decision seems further down the road.

After speaking with Ozz, I think the best small increment to deliver initially is a small change to the REST API to update/read config. Am going to try and think about the API design such that we don't have to rework it many times towards the desired behaviour.

A future step might be to use a Docker [bind mounts (https://docs.docker.com/engine/storage/bind-mounts/) to enable local JSON config idea, in addition to editing it via the CRUD API.

What I'm imagining a local JSON config could look like in future.

```json
{
    // ... other global config
    "workspaces": {
        "default": {
            "system_prompt": "",
            "muxing_rules": [
                {
                    "provider_name": "provider1",
                    "provider_id": "provider1_id",
                    "model": "model1",
                    "matcher_type": "catch_all",
                    "matcher": null
                },
                {
                    "provider_name": "provider2",
                    "provider_id": "provider2_id",
                    "model": "model2",
                    "matcher_type": "filename_match",
                    "matcher": "*.py"
                }
            ]
            // ... other workspace config
        },
        "foo": {
            "system_prompt": "",
            "muxing_rules": [
                {
                    "provider_name": "provider3",
                    "provider_id": "provider3_id",
                    "model": "model3",
                    "matcher_type": "request_type_match",
                    "matcher": "chat"
                }
            ]
            // ... other workspace config
        }
    }
}
[...]

PR #1107, opened 2025-02-19
feat: initial work on endpoints for creating/updating

alex-mcgovern:
Related to #1067 

- Modifies the existing POST `/workspaces`
  - it exclusively handles workspace creation — the ability to rename via this endpoint is removed
  - it can handle custom instructions & muxes configuration as part of the same request
- Adds a PUT `/workspaces/:workspace_name`
  - this is how you rename a workspace now
  - this can also handle updating custom instructions & muxes configuration

**ToDo**:
- GET `/workspaces/:workspace_name` or maybe (`/workspaces/:workspace_name`, I'm not sure yet) to get the configuration for a workspace

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -249,19 +249,12 @@ async def activate_workspace(request: v1_models.ActivateWorkspaceRequest, status
 @v1.post("/workspaces", tags=["Workspaces"], generate_unique_id_function=uniq_name, status_code=201)
 async def create_workspace(
     request: v1_models.CreateOrRenameWorkspaceRequest,
-) -> v1_models.Workspace:
+) -> v1_models.FullWorkspace:
     """Create a new workspace."""
-    if request.rename_to is not None:
-        return await rename_workspace(request)
-    return await create_new_workspace(request)
-
-
-async def create_new_workspace(
-    request: v1_models.CreateOrRenameWorkspaceRequest,
-) -> v1_models.Workspace:
-    # Input validation is done in the model
     try:
-        _ = await wscrud.add_workspace(request.name)
+        _ = await wscrud.add_workspace(
+            request.name, request.config.custom_instructions, request.config.muxing_rules
+        )
     except AlreadyExistsError:
         raise HTTPException(status_code=409, detail="Workspace already exists")
     except ValidationError:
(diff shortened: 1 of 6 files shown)


src/codegate/api/v1.py
```diff
@@ -280,15 +273,26 @@ async def create_new_workspace(
     return v1_models.Workspace(name=request.name, is_active=False)
```
alex-mcgovern, line 273: ToDo: return `FullWorkspace`

tests/api/test_v1_workspaces.py
alex-mcgovern, line 1: Naive approach to testing, given that we don't appear to spin up a DB during test runs anywhere that I have discovered so far — I think this could be a great addition, but open to alternatives.

src/codegate/muxing/models.py
```diff
@@ -41,7 +41,7 @@ class MuxRule(pydantic.BaseModel):
     def try_from_db_model(cls, db_model: DbMuxRule) -> "MuxRule":
         try:
             return cls(
-                provider_name=db_model.provider_endpoint_name,
+                provider_name=None,
```
JAORMX: why is this needed?

alex-mcgovern: It isn't needed and is me cleaning up a derp. This method was overwritten by a similar one in a commit Alejandro merged later anyway. 

Reviewing commit-by-commit might not be best as there is some scrappy incremental stuff in older commits.

JAORMX: gotcha, thanks!

src/codegate/workspaces/crud.py
```diff
@@ -31,34 +35,71 @@ class WorkspaceMuxRuleDoesNotExistError(WorkspaceCrudError):
...
+                if muxing_rules:
+                    mux_rules = await self.set_muxes(new_workspace_name, muxing_rules)
+
+                await transaction.commit()
+                return workspace_created, mux_rules
+            except WorkspaceNameAlreadyInUseError as e:
+                await transaction.rollback()
+                raise e
+            except Exception as e:
+                await transaction.rollback()
```
JAORMX, line 85: do you need these rollbacks? Isn't that already handled in the `DBTransaction` `__aexit__` function?

src/codegate/workspaces/crud.py
```diff
@@ -70,15 +111,42 @@ async def rename_workspace(
...
+                if muxing_rules:
+                    mux_rules = await self.set_muxes(new_workspace_name, muxing_rules)
+
+                await transaction.commit()
+                return workspace_renamed, mux_rules
+            except (WorkspaceNameAlreadyInUseError, WorkspaceDoesNotExistError) as e:
+                await transaction.rollback()
+                raise e
+            except Exception as e:
+                await transaction.rollback()
```
JAORMX, line 148: same comment about rollbacks

src/codegate/db/connection.py
```diff
@@ -993,6 +994,33 @@ async def get_distance_to_persona(
...
+            bind=DbCodeGate()._async_db_engine,
+            class_=AsyncSession,
+            expire_on_commit=False,
+        )()
+        await self._session.begin()
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        if exc_type:
+            await self._session.rollback()
```
JAORMX, line 1012: do you want to "re-raise" here?

src/codegate/workspaces/crud.py
alex-mcgovern: Will take a look, ty

src/codegate/db/connection.py
alex-mcgovern: that sounds wise

JAORMX (approved):

Issue #1060, opened 2025-02-14
[Task]: Adapt existing rules to work with Chat and FIM


aponcedeleonch:
### Description

We currently have a rule to differentiate between chat and FIM and a different one to differentiate between filenames. With the introduction of this feature we should join both rules into a single one.

PR #1112, opened 2025-02-19
Update the muxing rules to v3

aponcedeleonch:
Closes: #1060

Right now the muxing rules are designed to catch globally FIM or Chat requests. This PR extends its functionality to be able to match per file and request, i.e. this PR enables
- Chat request of main.py -> model 1
- FIM request of main.py -> model 2
- Any type of v1.py -> model 3

### DIFF
--- a/src/codegate/muxing/rulematcher.py
+++ b/src/codegate/muxing/rulematcher.py
@@ -18,6 +18,12 @@ _muxrules_sgtn = None
 _singleton_lock = Lock()
 
 
+class MuxMatchingError(Exception):
+    """An exception for muxing matching errors."""
+
+    pass
+
+
 async def get_muxing_rules_registry():
     """Returns a singleton instance of the muxing rules registry."""
 
@@ -48,9 +54,9 @@ class ModelRoute:
 class MuxingRuleMatcher(ABC):
     """Base class for matching muxing rules."""
 
-    def __init__(self, route: ModelRoute, matcher_blob: str):
+    def __init__(self, route: ModelRoute, mux_rule: mux_models.MuxRule):
         self._route = route
-        self._matcher_blob = matcher_blob
+        self._mux_rule = mux_rule
 
     @abstractmethod
     def match(self, thing_to_match: mux_models.ThingToMatchMux) -> bool:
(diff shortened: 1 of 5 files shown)


aponcedeleonch:
Marking as draft because this would break the FE. We need to coordinate to merge

lukehinds:
API Version 3?

peppescg:
on it https://github.com/stacklok/codegate-ui/issues/319 👨‍💻

src/codegate/muxing/rulematcher.py
```diff
@@ -67,32 +73,24 @@ class MuxingMatcherFactory:
...
+    def create(db_mux_rule: db_models.MuxRule, route: ModelRoute) -> MuxingRuleMatcher:
         """Create a muxing matcher for the given endpoint and model."""
 
         factory: Dict[mux_models.MuxMatcherType, MuxingRuleMatcher] = {
-            mux_models.MuxMatcherType.catch_all: CatchAllMuxingRuleMatcher,
-            mux_models.MuxMatcherType.filename_match: FileMuxingRuleMatcher,
-            mux_models.MuxMatcherType.request_type_match: RequestTypeMuxingRuleMatcher,
+            mux_models.MuxMatcherType.catch_all: RequestTypeAndFileMuxingRuleMatcher,
+            mux_models.MuxMatcherType.fim: RequestTypeAndFileMuxingRuleMatcher,
+            mux_models.MuxMatcherType.chat: RequestTypeAndFileMuxingRuleMatcher,
```
JAORMX, line 82: What happens to `filename_match` now?

JAORMX:
uhm... I'd be careful with having a v3 that's not backwards compatible without more usecases. This already starts looking too optimized for fim/chat in a way that it might not cater for other rule types.

src/codegate/muxing/rulematcher.py
aponcedeleonch: Now, we will have a single class to match `RequestTypeAndFileMuxingRuleMatcher`. It comprises both the logic of `filename_match` and `request_type` that we had before. Basically `RequestTypeAndFileMuxingRuleMatcher` will check if there is a filename present in the request and the type of request. Based on both parameters it will match and route the request to an LLM

aponcedeleonch:
> This already starts looking too optimized for fim/chat in a way that it might not cater for other rule types.

As I see it, it will be optimized for the request type. Right now we can only distinguish between 2 request types, fim or chat. Hence the types we have. In the future we would be able to match more request types, e.g. request type: Persona based

JAORMX (approved):

PR #1113, opened 2025-02-19
Copilot FIM doesn't always set model

jhrozek:
We had such fallback code in main, but removed it for some reason in the remove-litellm branch. I hit an issue today where the model was not set.

Let's default to `gpt-4o-mini` to be on the safe side.


jhrozek:
@blkt 👀

blkt (approved):

PR #1115, opened 2025-02-19
Handle OpenAI-like requests with prompt but no messages

jhrozek:
Some OpenAI-like requests, typically FIM requests don't set any messages, but instead set the `prompt` field. Let's add a wrapper Message-like structure with the right API so that the caller can use it to reach to the ChatCompletionMessage and set the prompt when handling FIM.

### DIFF
--- a/src/codegate/types/openai/_request_models.py
+++ b/src/codegate/types/openai/_request_models.py
@@ -277,6 +277,24 @@ class FunctionMessage(pydantic.BaseModel):
         yield self
 
 
+class PromptWrapperMessage(pydantic.BaseModel):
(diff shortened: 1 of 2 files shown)


jhrozek:
@blkt please treat this more like a draft. The PR works fine, but i'm not sure especially about handling of prompt vs. messages in the other methods of ChatCompletionRequest. Let's discuss.

blkt:
You're right, LLMs like llamacpp use OpenAI's legacy endpoints and request formats, I added the classes yesterday.

Let's discuss the integration later today.

blkt:
TL;DR: we'll use the `Legacy*` classes.

Issue #929, opened 2025-02-05
Refactor redaction mechanism to a single interface


lukehinds:
### Description

Right now we have secrets scanning , which encrypts matched tokens, secrets etc using AES CGM. This in turn has a session manager to track the mapping of secrets to encrypted strings. In hindsight this was over engineered, we really don't need this level of encryption. It's fine to just mask / redact the string, as the LLM can never determine what the original content was. This then leaves the session manager, again encryption does not add much here, as access to the local machine would result in access to the keys within the session manager.

For PII protections I took a different approach, which is to redact using UUIDs. UUIDs work well, they make a nice unique mappable element, have much less processing overhead.

With this in mind, I am repurposing this issue to have a single interface which accepts a string (secret, PII or user nominated string) and returns a UUID. A second interface can then be called, which presents a UUID and is returned with the unredacted metadata (secrets, PII etc). We also then end up with a single session manager.

We can then add context aware sanitation and refactor PII and secrets scanning to using a single interface reaction factory

PR #1123, opened 2025-02-20
feat: remove not needed encryption of secrets

yrobla:
Instead use an uuid generator as we do for pii, and reuse same session store mechanism

Closes: #929


src/codegate/pipeline/base.py
```diff
@@ -30,16 +30,8 @@ def secure_cleanup(self):
...
-
         self.manager.cleanup_session(self.session_id)
         self.session_id = ""
-
-        # Securely wipe the API key using the same method as secrets manager
-        if self.api_key is not None:
-            api_key_bytes = bytearray(self.api_key.encode())
-            self.manager.crypto.wipe_bytearray(api_key_bytes)
-            self.api_key = None
-
```
blkt, line 42: *question:* I understand this was removed because `self.manager` is not a `SecretsManager` object anymore, but shouldn't we wipe `self.api_key` anyway?

src/codegate/pipeline/sensitive_data/manager.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def to_json(self) -> str:
+        """Serializes the object to a JSON string."""
+        return json.dumps({key: value for key, value in vars(self).items() if value is not None})
+
+    @staticmethod
+    def from_json(data: str) -> "SensitiveData":
+        """Deserializes from a JSON string to a SensitiveData object."""
+        obj = json.loads(data)
+        return SensitiveData(obj["original"], obj.get("service"), obj.get("type"))
```
blkt, line 25: *nit:* I'm not a Python expert, but I believe we should be able to get this "for free" by extending `pydantic.BaseModel`.

src/codegate/pipeline/pii/pii.py
```diff
@@ -172,13 +271,13 @@ async def process_chunk(
         current_pos = 0
         result = []
         while current_pos < len(content):
-            start_idx = content.find("<", current_pos)
+            start_idx = content.find("#", current_pos)
```
blkt, line 274: *nit:* I would use the instance variables, I don't see much use for them otherwise.
```suggestion
            start_idx = content.find(self.marker_start, current_pos)
```

src/codegate/pipeline/pii/pii.py
```diff
@@ -172,13 +271,13 @@ async def process_chunk(
...
             if start_idx == -1:
                 # No more markers!, add remaining content
                 result.append(content[current_pos:])
                 break
 
-            end_idx = content.find(">", start_idx)
+            end_idx = content.find("#", start_idx + 1)
```
blkt, line 280: *nit:* I would use the instance variables, I don't see much use for them otherwise.
```suggestion
            end_idx = content.find(self.marker_end, start_idx + 1)
```

src/codegate/pipeline/pii/pii.py
```diff
@@ -75,33 +141,39 @@ async def process(
         total_pii_found = 0
         all_pii_details: List[Dict[str, Any]] = []
         last_redacted_text = ""
+        session_id = context.session_id if hasattr(context, "session_id") else str(uuid.uuid4())
```
blkt, line 144: *nit:* this looks like defensive code to me, can we remove it? I'd rather fix this then silently ignore a wrong argument.

blkt (commented):
Left a few comments.

src/codegate/pipeline/pii/pii.py
blkt: It seems to me I messed up this diff, but you get the idea.

src/codegate/pipeline/base.py
jhrozek: I think we might want to remove the `api_key` and its uses from the pipeline altogether. This used to be used when the pipeline had the ability to do "side calls" to LLMs to e.g. ask to extract packages from a snippet. but this was slow and replaced in favor of using tree-sitter.

As and advantage, the api_key will not leave the request->completion and will not be stored.

src/codegate/pipeline/sensitive_data/manager.py
```diff
@@ -0,0 +1,59 @@
...
+import structlog
+from codegate.pipeline.sensitive_data.session_store import SessionStore
+
+logger = structlog.get_logger("codegate")
+
+
+class SensitiveData:
+    """Represents sensitive data with additional metadata."""
+
+    def __init__(self, original: str, service: Optional[str], type: Optional[str]):
```
jhrozek, line 12: it might be nice to have an enum with the services instead of copying strings over. Or a constant in the modules if we want to keep a string (e.g. SENSITIVE_DATA_SERVICE="pii" in the PII pipeline and SENSITIVE_DATA_SERVICE="secrets" in the secrets pipeline)

jhrozek: Oh sorry I just re-read the rest and it seems like we are using a string in the secrets module that corresponds to the secret kind

jhrozek (approved):

PR #1137, opened 2025-02-20
Add a copilot-specific FIM class instead of adding extra fields to Le…

jhrozek:
…gacyCompletionRequest

Adding the copilot specific fields had the unintended side-effect of passing them to completion functions when we passed them as `**kwargs`.

In particular, this was breaking llamacpp FIM.

Instead, let's create another, copilot specific class that inherits from LegacyCompletionRequest.


jhrozek:
@blkt sorry about that, luckily we caught the regression before merging

PR #1158, opened 2025-02-25
Readme requires update

lukehinds:
The readme has falling behind and it not showcasing work such as workspaces and muxing, it also introduces that CodeGate is far more then just a security protection system.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,15 +21,15 @@
 
 ---
 
-# CodeGate: secure AI code generation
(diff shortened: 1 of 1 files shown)


yrobla:
@lukehinds this fix in README has conflicts with just the previous one you wrote... which one shall take preference?

lukehinds:
Ah, the other should not have merged.

Issue #1055, opened 2025-02-14
[Idea]: Persona-Based Muxing


aponcedeleonch:
Enhance CodeGate’s muxing functionality to support user-defined “personas,” allowing the system to classify incoming requests based on a persona and then route them to an LLM chosen by the user. For instance, a “Frontend React Expert” persona might be manually mapped to a favorite advanced LLM, while a “Backend Microservices Guru” persona could be routed to a lightweight local model—entirely at the user’s discretion.

### Why Is This Feature Important?

1. **Fine-Grained Control**. Users maintain complete authority over which LLM handles requests for a given persona (e.g., “Frontend React Expert” → Model X, “Backend Microservices Guru” → Model Y).
2. **Better Alignment With Developer Expertise**. By defining personas that capture specific skill sets or roles within a project, responses can be more targeted and relevant to the given domain, without sacrificing user choice in model selection.
3. **Resource and Cost Efficiency**. Users decide exactly when to employ advanced or specialized models, and when to default to smaller, cost-effective ones, based on the persona’s needs.

#### Persona Definitions

Store personas in CodeGate configuration (similar to how CodeGate currently handles different providers).

Examples:
- Frontend React Expert: Focuses on UI and React-specific queries.
- Backend Microservices Guru: Focuses on scalability, architecture, and performance.

#### Local LLM Classifier

A small, local model quickly inspects incoming prompts to determine which persona best fits.
Example: "How do I optimize state management in my React app?" → Frontend React Expert.

#### User-Defined LLM Routing

After classification, CodeGate routes requests to the LLM the user has configured for that persona.
Example: Frontend React Expert → [User-selected advanced model].
Users can easily update which LLM is tied to each persona at any time.

### Challenges & Considerations

1. **Classifier Accuracy**. Ensuring the local LLM correctly identifies the right persona. Misclassifications could lead to irrelevant or suboptimal answers—even if the correct LLM is specified.
2. **Performance & Latency**. Running a local model for classification adds a small overhead. Must be optimized to avoid bottlenecks in large-scale or rapid-fire scenarios. An alternative would be to use one of the user-defined providers. Although this would mean that CodeGate is going to consume more of the user's tokens which might not be expected and cause a bad impression.
3. **User Experience**. Providing a clear interface or config structure for defining personas and selecting their corresponding LLMs. Ensuring that changes to persona-LLM mappings are intuitive and quick to implement.
[...]

Issue #873, opened 2025-01-31
[Story]: Set custom instructions per filetype/model mux


danbarr:
### Description

Once per-filetype muxing is introduced into workspaces, it would be interesting to allow custom instructions at that level, in addition to or instead of instructions at the workspace level.

For example, docs/writing-specific instructions for *.md, language for framework-specific instructions for *.js vs. *.py etc.

Could be especially useful in monorepos or multi-language/multi-technology projects.

### Additional Context

/cc @aponcedeleonch, based on our quick brainstorm today

PR #1192, opened 2025-03-03
Added a class which performs semantic routing

aponcedeleonch:
Related to: #1055

For the current implementation of muxing we only need to match a single Persona at a time. For example:
1. mux1 -> persona Architect -> openai o1
2. mux2 -> catch all -> openai gpt4o

In the above case we would only need to know if the request matches the persona `Architect`. It's not needed to match any extra personas even if they exist in DB.

This PR introduces what's necessary to do the above without actually wiring in muxing rules. The PR:
- Creates the persona table in DB
- Adds methods to write and read to the new persona table
- Implements a function to check if a query matches to the specified persona

To check more about the personas and the queries please check the unit tests

### DIFF
--- a/migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
+++ b/migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
@@ -1,0 +1,50 @@
+"""add persona table
+
+Revision ID: 02b710eda156
+Revises: 5e5cd2288147
+Create Date: 2025-03-03 10:08:16.206617+00:00
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision: str = "02b710eda156"
+down_revision: Union[str, None] = "5e5cd2288147"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    # Begin transaction
+    op.execute("BEGIN TRANSACTION;")
+
+    op.execute(
+        """
+            CREATE TABLE IF NOT EXISTS personas (
+                id TEXT PRIMARY KEY,  -- UUID stored as TEXT
+                name TEXT NOT NULL UNIQUE,
+                description TEXT NOT NULL,
+                description_embedding BLOB NOT NULL
+            );
+            """
+    )
+
+    # Finish transaction
+    op.execute("COMMIT;")
+
+
+def downgrade() -> None:
+    # Begin transaction
+    op.execute("BEGIN TRANSACTION;")
+
+    op.execute(
+        """
+        DROP TABLE personas;
+        """
+    )
+
+    # Finish transaction
+    op.execute("COMMIT;")
--- a/src/codegate/config.py
+++ b/src/codegate/config.py
@@ -57,7 +57,9 @@ class Config:
     force_certs: bool = False
 
     max_fim_hash_lifetime: int = 60 * 5  # Time in seconds. Default is 5 minutes.
-    persona_threshold = 0.75  # Min value is 0 (max similarity), max value is 2 (orthogonal)
+    # Min value is 0 (max similarity), max value is 2 (orthogonal)
+    # The value 0.75 was found through experimentation. See /tests/muxing/test_semantic_router.py
+    persona_threshold = 0.75
 
     # Provider URLs with defaults
     provider_urls: Dict[str, str] = field(default_factory=lambda: DEFAULT_PROVIDER_URLS.copy())
--- a/src/codegate/config.py
+++ b/src/codegate/config.py
@@ -57,6 +57,7 @@ class Config:
     force_certs: bool = False
 
     max_fim_hash_lifetime: int = 60 * 5  # Time in seconds. Default is 5 minutes.
+    persona_threshold = 0.75  # Min value is 0 (max similarity), max value is 2 (orthogonal)
 
     # Provider URLs with defaults
     provider_urls: Dict[str, str] = field(default_factory=lambda: DEFAULT_PROVIDER_URLS.copy())
--- a/src/codegate/db/connection.py
+++ b/src/codegate/db/connection.py
@@ -560,7 +560,7 @@ class DbRecorder(DbCodeGate):
         )
 
         try:
-            # For Pydantic we conver the numpy array to a string when serializing.
+            # For Pydantic we convert the numpy array to string when serializing with .model_dumpy()
             # We need to convert it back to a numpy array before inserting it into the DB.
             persona_dict = persona.model_dump()
             persona_dict["description_embedding"] = persona.description_embedding
@@ -615,17 +615,19 @@ class DbReader(DbCodeGate):
                     raise e
                 return None
 
-    async def _exec_vec_db_query(
-        self, sql_command: str, conditions: dict
-    ) -> Optional[CursorResult]:
+    async def _exec_vec_db_query_to_pydantic(
+        self, sql_command: str, conditions: dict, model_type: Type[BaseModel]
+    ) -> List[BaseModel]:
         """
         Execute a query on the vector database. This is a separate connection to the SQLite
         database that has the vector extension loaded.
         """
         conn = self._get_vec_db_connection()
+        conn.row_factory = sqlite3.Row
         cursor = conn.cursor()
-        cursor.execute(sql_command, conditions)
-        return cursor
+        results = [model_type(**row) for row in cursor.execute(sql_command, conditions)]
+        conn.close()
+        return results
 
     async def get_prompts_with_output(self, workpace_id: str) -> List[GetPromptWithOutputsRow]:
         sql = text(
@@ -985,14 +987,10 @@ class DbReader(DbCodeGate):
             WHERE id = :id
         """
         conditions = {"id": persona_id, "query_embedding": query_embedding}
-        persona_distance_cursor = await self._exec_vec_db_query(sql, conditions)
-        persona_distance_raw = persona_distance_cursor.fetchone()
-        return PersonaDistance(
-            id=persona_distance_raw[0],
-            name=persona_distance_raw[1],
-            description=persona_distance_raw[2],
-            distance=persona_distance_raw[3],
+        persona_distance = await self._exec_vec_db_query_to_pydantic(
+            sql, conditions, PersonaDistance
         )
+        return persona_distance[0]
 
 
 def init_db_sync(db_path: Optional[str] = None):
--- a/src/codegate/db/connection.py
+++ b/src/codegate/db/connection.py
@@ -1,9 +1,12 @@
 import asyncio
 import json
+import sqlite3
 import uuid
 from pathlib import Path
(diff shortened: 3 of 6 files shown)


migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
```diff
@@ -0,0 +1,50 @@
...
+
+    op.execute(
+        """
+            CREATE TABLE IF NOT EXISTS personas (
+                id TEXT PRIMARY KEY,  -- UUID stored as TEXT
+                name TEXT NOT NULL UNIQUE,
+                description TEXT NOT NULL,
+                description_embedding BLOB NOT NULL
+            );
+            """
```
JAORMX, line 32: I think it's a good idea to have personas not be namespaced within a workspace since this allows us to share personas between workspaces. Do you think we should also add a namespaced persona concept? This is not a blocker and if we decide a namespaced persona makes sense, this can be left as a TODO for another PR.

src/codegate/config.py
```diff
@@ -57,6 +57,7 @@ class Config:
     force_certs: bool = False
 
     max_fim_hash_lifetime: int = 60 * 5  # Time in seconds. Default is 5 minutes.
+    persona_threshold = 0.75  # Min value is 0 (max similarity), max value is 2 (orthogonal)
```
JAORMX, line 60: Can you add also a comment in regards to why the `0.75` value was chosen?

src/codegate/db/connection.py
```diff
@@ -523,6 +545,30 @@ async def add_mux(self, mux: MuxRule) -> MuxRule:
...
+        """
+        sql = text(
+            """
+            INSERT INTO personas (id, name, description, description_embedding)
+            VALUES (:id, :name, :description, :description_embedding)
+            """
+        )
+
+        try:
+            # For Pydantic we conver the numpy array to a string when serializing.
```
JAORMX, line 563: `s/conver/convert/`

jhrozek:
Would it be possible to also add a special system prompt for the persona?

src/codegate/db/connection.py
```diff
@@ -569,6 +615,18 @@ async def _exec_select_conditions_to_pydantic(
...
 
+    async def _exec_vec_db_query(
+        self, sql_command: str, conditions: dict
+    ) -> Optional[CursorResult]:
+        """
+        Execute a query on the vector database. This is a separate connection to the SQLite
+        database that has the vector extension loaded.
+        """
+        conn = self._get_vec_db_connection()
+        cursor = conn.cursor()
```
JAORMX, line 626: You're currently not closing the cursor which is an issue, can you fix that? You probably want to use the `with conn.cursor() as cursor:` syntax instead so closing would happen automagically.

src/codegate/db/models.py
```diff
@@ -240,3 +241,39 @@ class MuxRule(BaseModel):
     priority: int
     created_at: Optional[datetime.datetime] = None
     updated_at: Optional[datetime.datetime] = None
+
+
+# Pydantic doesn't support numpy arrays out of the box. Defining a custom type
+# Reference: https://github.com/pydantic/pydantic/issues/7017
+def nd_array_custom_before_validator(x):
+    # custome before validation logic
```
JAORMX, line 249: you might want to reclarify this comment.

src/codegate/db/models.py
```diff
@@ -240,3 +241,39 @@ class MuxRule(BaseModel):
...
+
+class PersonaEmbedding(Persona):
+    description_embedding: NdArray  # sqlite-vec will handle numpy arrays directly
+
+    # Part of the workaround to allow numpy arrays in pydantic models
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+
+class PersonaDistance(Persona):
+    distance: float
```
JAORMX, line 279: Can you add more documentation on this distance? While the SQL query suggests this will end up as the cosign distance between the description and the embedding vectors, one has to reach out to another file to figure this out which is not great for maintenance.

src/codegate/muxing/semantic_router.py
```diff
@@ -0,0 +1,129 @@
...
+class SemanticRouter:
+
+    def __init__(self):
+        self._inference_engine = LlamaCppInferenceEngine()
+        conf = Config.get_config()
+        self._embeddings_model = f"{conf.model_base_path}/{conf.embedding_model}"
+        self._n_gpu = conf.chat_model_n_gpu_layers
+        self._persona_threshold = conf.persona_threshold
+        self._db_recorder = DbRecorder()
+        self._db_reader = DbReader()
```
JAORMX, line 29: Not for this Pull Request: I'm starting to think DbRecorder should inherit from DBReader so we wouldn't need to call an explicit object for each particular operation. Having two explicit objects makes sense in a distributed system where you might have read replicas, but it's probably overkill for this single container system.

src/codegate/muxing/semantic_router.py
```diff
@@ -0,0 +1,129 @@
...
+
+        # Normalize numbers (replace with placeholder)
+        text = re.sub(r"\b\d+\.\d+\b", " NUM ", text)  # Decimal numbers
+        text = re.sub(r"\b\d+\b", " NUM ", text)  # Integer numbers
+
+        # Replace punctuation with spaces (keeping apostrophes for contractions)
+        text = re.sub(r"[^\w\s\']", " ", text)
+
+        # Normalize whitespace (replace multiple spaces with a single space)
+        text = re.sub(r"\s+", " ", text)
```
JAORMX, line 77: optimization: pre-declare and pre-compile each regular expression in the constructor or even globally so this function would only need to evaluate the regex as opposed to compile+evaluate.

src/codegate/muxing/semantic_router.py
```diff
@@ -0,0 +1,129 @@
...
+        """
+        Helper function to embed text using the inference engine.
+        """
+        cleaned_text = self._clean_text_for_embedding(text)
+        # .embed returns a list of embeddings
+        embed_list = await self._inference_engine.embed(
+            self._embeddings_model, [cleaned_text], n_gpu_layers=self._n_gpu
+        )
+        # Use only the first entry in the list and make sure we have the appropriate type
+        logger.debug("Text embedded in semantic routing", text=cleaned_text[:100])
```
JAORMX, line 94: nit: is `100` characters overkill for the debug log? would 20 be enough?

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
```diff
@@ -0,0 +1,50 @@
...
+
+def upgrade() -> None:
+    # Begin transaction
+    op.execute("BEGIN TRANSACTION;")
+
+    op.execute(
+        """
+            CREATE TABLE IF NOT EXISTS personas (
+                id TEXT PRIMARY KEY,  -- UUID stored as TEXT
+                name TEXT NOT NULL UNIQUE,
```
JAORMX, line 28: I think we're missing a unique index for `name` here, are we? At least the code below tries to catch an integrity error when inserting a persona, which will only trigger if we have a unique index.

aponcedeleonch:
@jhrozek that's a nice suggestion. Yes, we can do that. Actually it would be nice to add a special system prompt for each of our muxing rules. Since personas is a muxing rule they would have a system prompt. The idea is tracked here: #873

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
```diff
@@ -0,0 +1,50 @@
...
+                description TEXT NOT NULL,
```
JAORMX, line 29: Do we want to make descriptions unique as well? If someone adds two similar descriptions, it would be very hard for the matcher to work properly. Perhaps enforcing uniqueness is the way to go as a first step, and in a further iteration we could check for description similarity. wdyt?

JAORMX (changes requested):
requesting changes mainly on the cursor closure.

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
aponcedeleonch: The keyword `UNIQUE` at the end of the line is enough to create the unique index for `name`

src/codegate/db/connection.py
aponcedeleonch: thanks, fixed

src/codegate/config.py
aponcedeleonch: yes, clarifiied :)

src/codegate/db/connection.py
aponcedeleonch: Nice catch! This is an `sqlite3` connection though, not an SQLAlchemy one. Hence, there's no context management, i.e. `with`. I did made changes to make sure the connection is closed though. It's not an SQLAlchemy because I haven't looked if we can load extensions with it. I at least know that with `SQLAlchemy + aiosqlite` engine I couldn't.

src/codegate/db/models.py
aponcedeleonch: Done! Let me know if more clarification is needed

src/codegate/db/models.py
aponcedeleonch: Yes, nice suggestion. I added more documentation

src/codegate/muxing/semantic_router.py
aponcedeleonch: I went with 50, 20 seems like too little. It's raw text so we should be able to know what chunk of text we embed

src/codegate/muxing/semantic_router.py
aponcedeleonch: You're right. Changed them :)

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
aponcedeleonch: That's a really nice suggestion! But actually making the descriptions unique won't cut it. If the difference is a single letter then we will accept the new description. What I will do is to check the cosine distance to the existing descriptions and only accept a new persona if it's sufficiently different. Will upload a commit soon

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
JAORMX: d'oh! totally missed that

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
aponcedeleonch: Uuhm <thinking>. Probably the concept of persona namespaces makes sense. Although right now I can't think on what would be the difference wrt. our workspaces. In other words, I like the idea but lack the use cases atm. Lets introduce when we need them

migrations/versions/2025_03_03_1008-02b710eda156_add_persona_table.py
JAORMX: cool beans

JAORMX (approved):
shipit :ship:

PR #1193, opened 2025-03-03
Use glob patterns in muxing rules

aponcedeleonch:
We were using a handcrafted logic to determine if it was an exact match or an extension match (sorry about that).

This PR replaces it for glob patterns


danbarr:
With this change, will an empty input still match all files?

JAORMX (approved):

danbarr (approved):

Issue #1061, opened 2025-02-14
codegate cli doesn't work in copilot if context is selected


jhrozek:
Found this while porting the new pipeline code. I thought it was a bug in our code, but it must be a long-standing bug in the copilot provider.

alex-mcgovern:
+1 — came here to report this. I suspect this may be related to the recent copilot release with the agentic/compose functionality — but unsure.

https://github.com/user-attachments/assets/93f0cd8e-e59f-408a-b285-7b518e5736f5

jhrozek:
To fix this, we need to add code to "src/codegate/pipeline/cli/cli.py as we alrady have for other clients, this time for copilot

danbarr:
This is also impacting the new avante.nvim integration.

danbarr:
I just upgraded to v0.1.28 which includes #1202 but this is still an issue for me

<img width="423" alt="Image" src="https://github.com/user-attachments/assets/0007f6b1-0012-4e55-9fbf-a59b0c02ae0b" />

PR #1202, opened 2025-03-04
Strip attachments from copilot messages when handling the CLI commands

jhrozek:
Copilot sends files in the context in the following format:
```
<attachment>
file
<attachment>
user query
```

Fixes: #1061

### DIFF
--- a/src/codegate/pipeline/cli/cli.py
+++ b/src/codegate/pipeline/cli/cli.py
@@ -95,6 +95,25 @@ def _get_cli_from_continue(last_user_message_str: str) -> Optional[re.Match[str]
     return codegate_regex.match(last_user_message_str)
 
 
+def _get_cli_from_copilot(last_user_message_str: str) -> Optional[re.Match[str]]:
+    """
+    Process Copilot-specific CLI command format.
+
+    Copilot sends messages in the format:
+    <attachment>file contents</attachment>codegate command
+
+    Args:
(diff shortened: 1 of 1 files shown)


src/codegate/pipeline/cli/cli.py
```diff
@@ -94,6 +94,21 @@ def _get_cli_from_continue(last_user_message_str: str) -> Optional[re.Match[str]
...
+    Copilot sends messages in the format:
+    <attachment>file contents</attachment>codegate command
+
+    Args:
+        last_user_message_str (str): The message string from Copilot
+
+    Returns:
+        Optional[re.Match[str]]: A regex match object if command is found, None otherwise
+    """
+    cleaned_text = re.sub(r'<attachment>.*</attachment>', '', last_user_message_str, flags=re.DOTALL)
```
JAORMX, line 110: optimization: pre-declare and pre-compile each regular expression in the constructor or even globally so this function would only need to evaluate the regex as opposed to compile+evaluate.

JAORMX (dismissed):

src/codegate/pipeline/cli/cli.py
jhrozek: let's do it separately for all the regexes here. We should also move the client helpers to the client interface.

JAORMX (approved):

PR #1204, opened 2025-03-04
Disable suspicious commands for now

therealnb:
This is not quite the right place for the code check to go.

We are meeting to discuss this further. Currently it is running on all files.


JAORMX (approved):

therealnb:
Thanks @JAORMX !

PR #1215, opened 2025-03-04
Run make all to reformat code automatically

jhrozek:
I don't know why our lint CI step doesn't catch this (it did slap me the
other day for other reasons, so it is working), but running `make all`
produced a bunch of reformats.


JAORMX (approved):
lgtm

aponcedeleonch (approved):

Issue #1218, opened 2025-03-05
[Story]: Create Personas with descriptions


aponcedeleonch:
### Description

The descriptions should be embedded and stored in DB to then be able to perform semantic searches by persona

aponcedeleonch:
Persona creation in [#1192](https://github.com/stacklok/codegate/pull/1192)

Validation of description missing

PR #1225, opened 2025-03-05
Validate persona description is sufficiently different

aponcedeleonch:
Closes: #1218

Check if the description for a new persona is different enough from the existing personas descriptions. This is done to correctly differentiate between personas

### DIFF
--- a/src/codegate/muxing/semantic_router.py
+++ b/src/codegate/muxing/semantic_router.py
@@ -28,6 +28,10 @@ class PersonaDoesNotExistError(Exception):
     pass
 
 
+class PersonaSimilarDescriptionError(Exception):
+    pass
+
+
 class SemanticRouter:
 
     def __init__(self):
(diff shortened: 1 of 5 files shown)


src/codegate/muxing/semantic_router.py
```diff
@@ -105,12 +110,38 @@ async def _embed_text(self, text: str) -> np.ndarray:
...
+            return True
+
+        for persona_distance in persona_distances:
+            logger.info(
+                f"Persona description distance to {persona_distance.name}",
+                distance=persona_distance.distance,
+            )
+            # If the distance is less than the threshold, the persona description is too similar
+            if persona_distance.distance < self._persona_diff_desc_threshold:
+                return False
```
JAORMX, line 131: How expensive is it to parallelize this?

JAORMX (approved):

src/codegate/muxing/semantic_router.py
aponcedeleonch: I tried 2 approaches to get the distances:
1. Direct query to sqlite and get the distances
2. Query to get all the personas, then use `numpy` matrices operations to get the distance

The result of the experiment was that it didn't matter, both of them were practically onpar.

For this specific comparison for just checking the threshold probably makes no difference to parallelize it with matrices operations. I don't expect someone having 1000 different personas in their DB. If it happens then yes, we would need for optimization. Probably with a sensible amount of personas (<10) really makes no difference

JAORMX: Let's take an extreme but not unreasonable example: 100 personas. Would we start seeing issues in this case?

Issue #1219, opened 2025-03-05
[Task]: CRUD for Personas


aponcedeleonch:
### Description

Create CRUD for Personas with their respective API endpoint

PR #1232, opened 2025-03-05
Created necessary methods for Persona CRUD

aponcedeleonch:
Closes: #1219

Some changes in this PR
- Renamed Semantic Router to PersonaManager: The motivation is that the only semantic routing we're doing is based on persona. So lets just call it that wat
- Created update and delete methods for Persona
- Added tests for the whole Persona CRUD

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -12,7 +12,12 @@ import codegate.muxing.models as mux_models
 from codegate import __version__
 from codegate.api import v1_models, v1_processing
 from codegate.db.connection import AlreadyExistsError, DbReader
-from codegate.db.models import AlertSeverity, WorkspaceWithModel
+from codegate.db.models import AlertSeverity, Persona, WorkspaceWithModel
+from codegate.muxing.persona import (
+    PersonaDoesNotExistError,
+    PersonaManager,
+    PersonaSimilarDescriptionError,
+)
 from codegate.providers import crud as provendcrud
 from codegate.workspaces import crud
 
@@ -21,6 +26,7 @@ logger = structlog.get_logger("codegate")
 v1 = APIRouter()
 wscrud = crud.WorkspaceCrud()
 pcrud = provendcrud.ProviderCrud()
+persona_manager = PersonaManager()
 
 # This is a singleton object
 dbreader = DbReader()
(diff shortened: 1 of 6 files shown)


peppescg (approved):
✈️

src/codegate/api/v1.py
```diff
@@ -665,3 +671,89 @@ async def get_workspace_token_usage(workspace_name: str) -> v1_models.TokenUsage
...
+
+
+@v1.get("/personas/{persona_name}", tags=["Personas"], generate_unique_id_function=uniq_name)
+async def get_persona(persona_name: str) -> Persona:
+    """Get a persona by name."""
+    try:
+        persona = await dbreader.get_persona_by_name(persona_name)
+        if not persona:
+            raise HTTPException(status_code=404, detail=f"Persona {persona_name} not found")
+        return persona
```
JAORMX, line 694: should we be catching the exception when a persona doesn't exist?

src/codegate/api/v1.py
```diff
@@ -665,3 +671,89 @@ async def get_workspace_token_usage(workspace_name: str) -> v1_models.TokenUsage
...
+    except Exception as e:
+        if isinstance(e, HTTPException):
```
JAORMX, line 696: Should the db reader be raising an HTTP exception? This sounds off

JAORMX (changes requested):

src/codegate/api/v1.py
```diff
@@ -665,3 +671,89 @@ async def get_workspace_token_usage(workspace_name: str) -> v1_models.TokenUsage
     except Exception:
         logger.exception("Error while getting messages")
         raise HTTPException(status_code=500, detail="Internal server error")
+
+
+@v1.get("/personas", tags=["Personas"], generate_unique_id_function=uniq_name)
+async def list_personas() -> List[Persona]:
+    """List all personas."""
+    try:
+        personas = await dbreader.get_all_personas()
```
JAORMX, line 680: why not leverage the persona manager here to keep the code consistent? Adding the dbreader here seems like an antipattern

src/codegate/api/v1.py
```diff
@@ -665,3 +671,89 @@ async def get_workspace_token_usage(workspace_name: str) -> v1_models.TokenUsage
...
+        return personas
+    except Exception:
+        logger.exception("Error while getting personas")
+        raise HTTPException(status_code=500, detail="Internal server error")
+
+
+@v1.get("/personas/{persona_name}", tags=["Personas"], generate_unique_id_function=uniq_name)
```
JAORMX, line 687: Can we use an annotated object to do input validation on `persona_name`? I'd like to restrict the characters that folks can use here to prevent vulnerabilities.

PR #1233, opened 2025-03-05
shareable workspaces pt. 2

alex-mcgovern:
Second part of https://github.com/stacklok/codegate/issues/1067 — this adds a new endpoint `GET /api/v1/workspaces/:workspace_name` that returns a full workspace config.

With this we have the prerequisites for a upload/download flow on a per-workspace basis.

These changes are integrated into the frontend in this PR: https://github.com/stacklok/codegate-ui/pull/379

- adds `provider_endpoint_name` and `provider_endpoint_type` to the `muxes` table
- changes `MuxRule` model used by the API to include `provider_type` and
  `provider_name` and removes `provider_id`
- creates a new `MuxRuleWithProviderId` model used internally for transporting
  mux rules
- all endpoints that interact with provider endpoints have been updated so that
  they don't require knowledge of the provider ID
- `GET /api/v1/workspaces/:provider_id` was already in use for listing providers in use by a workspace, but after speaking with @JAORMX we decided to move this elsewhere. You can achieve the same thing with `GET /api/v1/workspaces?provider_name={ID}` now instead
- there are a number of other breaking changes impacting the UI, but these are
  dealt with in a separate PR


migrations/versions/2025_03_06_1324-4b81c45b5da6_add_provider_endpoint_name_to_muxes.py
```diff
@@ -0,0 +1,100 @@
+"""add provider_endpoint_name to muxes
+
+Revision ID: 4b81c45b5da6
+Revises: 769f09b6d992
+Create Date: 2025-03-06 13:24:41.123857+00:00
+
+"""
+
```
aponcedeleonch, line 8: This migration seems really similar to the other one. Is it repeated?

migrations/versions/2025_03_07_0902-1ee1be2156f7_.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+
+from typing import Sequence, Union
+
+# revision identifiers, used by Alembic.
+revision: str = "1ee1be2156f7"
+down_revision: Union[str, None] = ("e4c05d7591a8", "4b81c45b5da6")
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
```
aponcedeleonch, line 16: If we have a single migration we don't need a merge migration

src/codegate/api/v1.py
```diff
@@ -36,6 +35,32 @@ def uniq_name(route: APIRoute):
     return f"v1_{route.name}"
 
 
+async def _add_provider_id_to_mux_rule(
+    mux_rule: mux_models.MuxRule,
+) -> mux_models.MuxRuleWithProviderId:
+    """
+    Convert a `MuxRule` to `MuxRuleWithProviderId` by looking up the provider ID.
+    Extracts provider name and type from the MuxRule and queries the database to get the ID.
+    """
```
aponcedeleonch, line 44: This function is not really an endpoint. Maybe it would be better to have it as a `classmethod` of `MuxRuleWithProviderId`?

src/codegate/api/v1.py
```diff
@@ -76,32 +101,36 @@ async def list_all_models_for_all_providers() -> List[v1_models.ModelByProvider]
...
-    provider_id: UUID,
+    provider_name: str,
 ) -> List[v1_models.ModelByProvider]:
     """List models by provider."""
 
     try:
-        return await pcrud.models_by_provider(provider_id)
+        provider = await pcrud.get_endpoint_by_name(provider_name)
+        if provider is None:
+            raise provendcrud.ProviderNotFoundError
```
aponcedeleonch, line 116: Lets move this validation inside `pcrud.get_endpoint_by_name(provider_name)` and just catch it here

src/codegate/api/v1.py
```diff
@@ -147,18 +176,21 @@ async def add_provider_endpoint(
...
-    provider_id: UUID,
+    provider_name: str,
     request: v1_models.ConfigureAuthMaterial,
 ):
     """Configure auth material for a provider."""
     try:
-        await pcrud.configure_auth_material(provider_id, request)
+        provider = await pcrud.get_endpoint_by_name(provider_name)
+        if provider is None:
+            raise provendcrud.ProviderNotFoundError
```
aponcedeleonch, line 192: Same as the last one. Lets move it inside `pcrud`

src/codegate/api/v1.py
```diff
@@ -322,6 +392,7 @@ async def update_workspace(
             ),
         )
     except crud.WorkspaceCrudError as e:
+        logger.debug(f"Could not update workspace: {e}")
```
aponcedeleonch, line 395: Make this `logger.exception` so we get in the logs the stacktrace

src/codegate/api/v1.py
alex-mcgovern: nice 👍🏻

migrations/versions/2025_03_07_0902-1ee1be2156f7_.py
alex-mcgovern: this was the result of `alembic merge heads` after I pulled from main — I think the head conflict was created when the installation table was added

migrations/versions/2025_03_06_1324-4b81c45b5da6_add_provider_endpoint_name_to_muxes.py
alex-mcgovern: there's 2, add `provider_endpoint_name` and `provider_endpoint_type` — I realised I needed both, and seemed cleaner to leave them separated, I can rewrite it as one though

aponcedeleonch: I think it's better if we have just 1 yes :)

migrations/versions/2025_03_07_0902-1ee1be2156f7_.py
aponcedeleonch: ooook, I think if we go with the single migration approach this extra migration won't be needed

migrations/versions/2025_03_06_1130-769f09b6d992_add_provider_endpoint_fields_to_muxes.py
```diff
@@ -0,0 +1,133 @@
...
+        ALTER TABLE muxes
+        ADD COLUMN provider_endpoint_type TEXT;
+        """
+    )
+    op.execute(
+        """
+        ALTER TABLE muxes
+        ADD COLUMN provider_endpoint_name TEXT;
+        """
+    )
```
JAORMX, line 36: Do we need these new references in the table? We could probably first just query the providers table to get the ID and then keep using the database schema as-is.

migrations/versions/2025_03_06_1130-769f09b6d992_add_provider_endpoint_fields_to_muxes.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # Delete mux rules where provider_endpoint_id doesn't match a provider in
```
alex-mcgovern, line 38: on delete cascade tirgger can simplify this

src/codegate/db/connection.py
```diff
@@ -1073,11 +1119,64 @@ async def get_provider_endpoint_by_name(self, provider_name: str) -> Optional[Pr
...
+
+        This makes the shareable workspace feature a little more robust.
+        """
+        # First try exact match on both name and type
         sql = text(
             """
             SELECT id, name, description, provider_type, endpoint, auth_type, created_at, updated_at
             FROM provider_endpoints
+            WHERE name = :name AND provider_type = :provider_type
+            LIMIT 1
```
alex-mcgovern, line 1142: not needed

src/codegate/workspaces/crud.py
```diff
@@ -254,8 +262,15 @@ async def soft_delete_workspace(self, workspace_name: str):
             raise WorkspaceCrudError(f"Error deleting workspace {workspace_name}")
 
         # Remove the muxes from the registry
-        mux_registry = await rulematcher.get_muxing_rules_registry()
-        await mux_registry.delete_ws_rules(workspace_name)
+        try:
+            mux_registry = await rulematcher.get_muxing_rules_registry()
+            rules = await mux_registry.get_ws_rules(workspace_name)
+            if rules:
+                await mux_registry.delete_ws_rules(workspace_name)
```
JAORMX, line 269: Can we handle the exception in case there are no rules to delete from inside `delete_ws_rules`? It would be great if we could just call it and it would not fail if there are no rules to delete. It still did the right thing.

JAORMX (changes requested):
The main thing is the db schema changes

src/codegate/workspaces/crud.py
alex-mcgovern: I moved the check inside the method so at least we're not holding the lock twice

```python
    async def delete_ws_rules(self, workspace_name: str) -> None:
        """Delete the rules for the given workspace."""
        async with self._lock:
            if workspace_name in self._ws_rules:
                del self._ws_rules[workspace_name]
```

src/codegate/db/connection.py
```diff
@@ -416,6 +416,16 @@ async def soft_delete_workspace(self, workspace: WorkspaceRow) -> Optional[Works
...
 
     async def hard_delete_workspace(self, workspace: WorkspaceRow) -> Optional[WorkspaceRow]:
+        # First delete associated muxes
+        sql_delete_muxes = text(
+            """
+            DELETE FROM muxes
+            WHERE workspace_id = :id
+            """
+        )
+        await self._execute_with_no_return(sql_delete_muxes, {"id": workspace.id})
```
JAORMX, line 426: Is this needed? Muxes should have an `ON DELETE CASCADE` that would clean them up if a workspace is hard-deleted.

lukehinds (approved):
Overall lgtm , do we need a db migration for the changes to db/models?

alex-mcgovern:
> Overall lgtm , do we need a db migration for the changes to db/models?

There were DB schema changes in a previous iteration, but we decided they were unnecessary, the only change to a DB model is this one:

```diff
class ProviderModel(BaseModel):
    provider_endpoint_id: str
+   provider_endpoint_type: str
    provider_endpoint_name: Optional[str] = None
    name: str
```

Which is populated via a JOIN — it's not the cleanest, but "it works"™️

JAORMX (approved):

PR #1234, opened 2025-03-05
Add `instance` table along with init code.

blkt:
This change adds an `instance` table containing the minimum set of details about codegate. We might want to add more details in the future.

The table must contain only a single record at any time. To guarantee that, a trigger is added that prevents inserts beyond the first record.

Finally, code performing the initialization is added to the `serve` command.

### DIFF
--- a/migrations/versions/2025_03_05_2126-e4c05d7591a8_add_installation_table.py
+++ b/migrations/versions/2025_03_05_2126-e4c05d7591a8_add_installation_table.py
@@ -1,0 +1,63 @@
+"""add installation table
+
+Revision ID: e4c05d7591a8
+Revises: 3ec2b4ab569c
+Create Date: 2025-03-05 21:26:19.034319+00:00
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = "e4c05d7591a8"
+down_revision: Union[str, None] = "3ec2b4ab569c"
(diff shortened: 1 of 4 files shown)


migrations/versions/2025_03_05_2126-e4c05d7591a8_add_installation_table.py
```diff
@@ -0,0 +1,63 @@
...
+        -- is fixed.
+
+        CREATE TRIGGER single_instance
+        BEFORE INSERT ON instance
+        WHEN (SELECT COUNT(*) FROM instance) >= 1
+        BEGIN
+          SELECT RAISE(FAIL, 'only one instance!');
+        END;
+        """
+    )
```
JAORMX, line 47: could we have a trigger that prevents deletion of the one row as well?

JAORMX (dismissed):

migrations/versions/2025_03_05_2126-e4c05d7591a8_add_installation_table.py
blkt: I would avoid adding a constraint on deletion, it becomes cumbersome to modify the table since AFAIK you cannot disable triggers temporarily in sqlite, which means you'd have to rely on `DROP TABLE/CREATE TABLE` to fix any sort of issue.

Additionally, it's hard to delete that record by mistake since there's no `delete_instance` routine in `DbRecorder` and initialization takes care of recreating it.

I'm inclined to let this as is, but I'm not religious about it, we can add it now or later anyway.

JAORMX (approved):

PR #1243, opened 2025-03-06
feat(vector-db): add cve_packages table

samuv:
This PR introduces a new table, cve_packages, to store package-version pairs that have at least one vulnerability classified as high or critical. The data is stored in the S3 bucket alongside malicious, deprecated, and archived data. The goal is to have this data ready for when we export the dependency list. The data is generated by the GitHub Action and is currently triggered manually.

Changes

Added a new table cve_packages with the following fields:
- 	name (TEXT, NOT NULL) – package name
- 	version (TEXT, NOT NULL) – package version
- 	type (TEXT, NOT NULL) – package type (e.g., npm, pypi)

Created indexes on the name, name:version, and name:version:type fields to optimize queries.
Modified import_packages.py to process and insert package versions with high/critical vulnerabilities into cve_packages.

### DIFF
--- a/.github/workflows/import_packages.yml
+++ b/.github/workflows/import_packages.yml
@@ -47,6 +47,7 @@ jobs:
         MALICIOUS_KEY=$(jq -r '.latest.malicious_packages' manifest.json)
         DEPRECATED_KEY=$(jq -r '.latest.deprecated_packages' manifest.json)
         ARCHIVED_KEY=$(jq -r '.latest.archived_packages' manifest.json)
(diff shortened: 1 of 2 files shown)


aponcedeleonch (approved):
Looks good! Should we calculate the embeddings for the packages with CVEs? Or is that left for a future PR?

samuv:
> Looks good! Should we calculate the embeddings for the packages with CVEs? Or is that left for a future PR?

for packages with CVEs, we realized that embeddings aren’t needed since it’s just a straightforward select query for package:version

Issue #1220, opened 2025-03-05
[Task]: Wire in Personas as muxing rules


aponcedeleonch:
### Description

Add a new muxing rule which allows the user to select a Persona as a muxing rule

PR #1244, opened 2025-03-06
Wire in Persona matching to muxing

aponcedeleonch:
Closes: #1220

This PR:
- Changes the way the queries are checked to determine if they match a persona description. Before we were using SQLite now we use Numpy
- Adds a new MuxMatcherType called `persona_description` that checks if the user messages in a request match a persona description


src/codegate/muxing/persona.py
```diff
@@ -217,21 +218,87 @@ async def delete_persona(self, persona_name: str) -> None:
...
+
+        Returns:
+            Weighted distances as a NumPy array
+        """
+        # Get array length
+        n = len(distances)
+
+        # Create positions array in reverse order (n-1, n-2, ..., 1, 0)
+        # This makes the last element have position 0
+        positions = np.arange(n - 1, -1, -1)
```
JAORMX, line 271: nit: Mind using named parameters here so folks don't have to look up the function declaration?

JAORMX (commented):
So far this makes sense to me. Time to test it!

src/codegate/muxing/persona.py
aponcedeleonch: sure, no problem. Changed it

lukehinds:
Hold on this , had discussion earlier and there is a change of plan

lukehinds (changes requested):
Hold on this , had discussion earlier and there is a change of plan

PR #1247, opened 2025-03-07
feat: update messages endpoint to return a conversation summary

yrobla:
Modify the messages endpoint to return just a conversationsummary, that will simplify the current queries.
Create a different endpoint that will return a list of conversations for a given prompt id

### DIFF
--- a/src/codegate/api/v1.py
+++ b/src/codegate/api/v1.py
@@ -420,7 +420,9 @@ async def get_workspace_alerts(workspace_name: str) -> List[Optional[v1_models.A
         raise HTTPException(status_code=500, detail="Internal server error")
 
     try:
-        alerts = await dbreader.get_alerts_by_workspace(ws.id, AlertSeverity.CRITICAL.value)
+        alerts = await dbreader.get_alerts_by_workspace_or_prompt_id(
+            workspace_id=ws.id, trigger_category=AlertSeverity.CRITICAL.value
+        )
         prompts_outputs = await dbreader.get_prompts_with_output(ws.id)
         return await v1_processing.parse_get_alert_conversation(alerts, prompts_outputs)
     except Exception:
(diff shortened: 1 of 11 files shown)


src/codegate/api/v1.py
```diff
@@ -469,19 +477,103 @@ async def get_workspace_messages(workspace_name: str) -> List[v1_models.Conversa
...
-                ws.id, AlertSeverity.CRITICAL.value
-            )
-        )
-        conversations, _ = await v1_processing.parse_messages_in_conversations(
-            prompts_with_output_alerts_usage
+    offset = (page - 1) * page_size
+    valid_conversations: List[v1_models.ConversationSummary] = []
+    fetched_prompts = 0
+
+    while len(valid_conversations) < page_size:
```
alex-mcgovern, line 484: we can get into an infinite loop if the page size is greater than the total, do we just need to break early?

```python
        prompts = await dbreader.get_prompts(
            ws.id,
            offset + fetched_prompts,
            batch_size,
            filter_by_ids,
            list([AlertSeverity.CRITICAL.value]),  # TODO: Configurable severity
            filter_by_alert_trigger_types,
        )

        if not prompts:
            break
```

src/codegate/api/v1.py
```diff
@@ -665,7 +757,7 @@ async def get_workspace_token_usage(workspace_name: str) -> v1_models.TokenUsage
         raise HTTPException(status_code=500, detail="Internal server error")
 
     try:
-        prompts_outputs = await dbreader.get_prompts_with_output(ws.id)
+        prompts_outputs = await dbreader.get_prompts_with_output(worskpace_id=ws.id)
```
alex-mcgovern, line 760: small typo was causing this endpoint to 500
```suggestion
        prompts_outputs = await dbreader.get_prompts_with_output(workspace_id=ws.id)
```

alex-mcgovern (approved):

PR #1248, opened 2025-03-07
Allow for dynamic workspace selection for muxing

JAORMX:
This allows for the muxing endpoint to detect the `X-CodeGate-Workspace`
header in requests. This will allow CodeGate to search for that
workspace and use it's routing configuration explicitly.

If the a workspace is requested and it does not exist, a warning will be
issued and the active workspace will be defaulted to.

The use case for this is for background agents to be able to select a
relevant workspace and not need to disrupt the user's IDE session.

Note that this is not a persistent session and merely a dynamic runtime
setting.


JAORMX:
So, you could have several workspaces with different rules and filters. This would allow an agent like Stitch to choose the workspace they want to take into use; so one wouldn't disrupt the user's IDE session if they have one open. Next week I'll do a rule matcher that selects from a "label" which will simply be an HTTP header. Then from stitch we can set the header per expert.

Issue #1250, opened 2025-03-07
codegate garbles output on main


jhrozek:
Found when debugging our llmlite-less branch, but turns out this bug is on main as well

<img width="408" alt="Image" src="https://github.com/user-attachments/assets/c018e9e2-6e31-490b-80ad-ca75ddf91350" />

jhrozek:
The bug is in the secret unredaction step.

PR #1261, opened 2025-03-10
Fix streaming output corruption with copilot

jhrozek:
- **Shortcut other steps if one holds the output** - We had a bug in the output pipeline where if one step returned [] meaning that the output chunk should be held off, all the current chunks would continue to run and potentially modify the context.
- **Shortcut buffered PII sooner, as soon as the buffer can't be a UUID** - Our PII refaction format is `#UUID#`. Our code was finding an opening #, then checking for a closing matching # or end of the output. For copilot, however, this meant that we were buffering the whole file, because the filename comes in this format:
```
# filepath: /Users/jakub/devel/demo-repo-python/app.py
```
This means we would keep searching for the closing hash which never
came. Instead, buffer only as long as the context between the hashes can
reasonably be a UUID.

Fixes: #1250


jhrozek:
@blkt I can rebase this atop replace-litellm if you prefer

JAORMX (dismissed):

jhrozek:
just rebased atop Rado's test fix

JAORMX (approved):

PR #1264, opened 2025-03-11
Use updates.codegate.ai for /versions instead of GitHub

dmjb:
Wire up CodeGate to use the new update service to determine the latest version. This is kept behind a config flag, but I have tested locally.

### DIFF
--- a/src/codegate/config.py
+++ b/src/codegate/config.py
@@ -59,6 +59,10 @@ class Config:
     server_key: str = "server.key"
     force_certs: bool = False
 
+    # Update configuration.
+    use_update_service: bool = False
+    update_service_url: str = "https://updates.codegate.ai/api/v1/version"
+
     max_fim_hash_lifetime: int = 60 * 5  # Time in seconds. Default is 5 minutes.
 
     # Min value is 0 (max similarity), max value is 2 (orthogonal)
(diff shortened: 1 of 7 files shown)


src/codegate/config.py
```diff
@@ -209,11 +215,17 @@ def from_env(cls) -> "Config":
             if "CODEGATE_SERVER_KEY" in os.environ:
                 config.server_key = os.environ["CODEGATE_SERVER_KEY"]
             if "CODEGATE_FORCE_CERTS" in os.environ:
-                config.force_certs = os.environ["CODEGATE_FORCE_CERTS"]
+                config.force_certs = cls.__bool_from_string(os.environ["CODEGATE_FORCE_CERTS"])
```
dmjb, line 218: Fix for a bug I spotted while working on this part of the code: `force_certs` expects a bool, but in this line we assign a string to the value. The likely consequence of this is that if someone explicitly sets this value to `false` in an env var, it will end up evaluating to `True` when treated as a boolean.

alex-mcgovern (approved):
Didn't spot anything weird — looks good to me, frontend still works.

Screenshot 2025-03-11 at 4 21 07 PM

Only question is do we need the config flag? Can we not just validate roughly on this branch and include in next release?

PR #1268, opened 2025-03-12
Run update call on recurring schedule

dmjb:
Call to the update service every four hours, and use the BE origin type. Print a warning level log message if an update is available.

This PR also introduces some refactoring from the previous PR:

1) Refactor the update client to be a singleton.
2) Set the instance ID once on application load.
3) Get rid of the feature flag - using the new service is now default.


src/codegate/updates/scheduled.py
```diff
@@ -0,0 +1,34 @@
...
+
+class ScheduledUpdateChecker(threading.Thread):
+    """
+    ScheduledUpdateChecker calls the UpdateClient on a recurring interval.
+    This is implemented as a separate thread to avoid blocking the main thread.
+    A dedicated scheduling library could have been used, but the requirements
+    are trivial, and a simple hand-rolled solution is sufficient.
+    """
+
+    def __init__(self, client: UpdateClient, interval_seconds: int = 14400):  # 4 hours in seconds
```
dmjb, line 20: Not sure if we want this interval to be externally configurable?

tests/test_server.py
```diff
@@ -75,18 +75,20 @@ def test_health_check(test_client: TestClient) -> None:
     assert response.json() == {"status": "healthy"}
 
 
-@patch("codegate.api.v1_processing.fetch_latest_version", return_value="foo")
-def test_version_endpoint(mock_fetch_latest_version, test_client: TestClient) -> None:
+@patch("codegate.api.v1._get_latest_version")
```
dmjb, line 78: Ideally this test would use a patched out UpdateClient instance, but it's difficult to patch out a singleton with the mocks. 

This sort of testing would be much easier if the V1 routes were defined in a class, and all dependencies were supplied via the constructor.

blkt (approved):

PR #1269, opened 2025-03-12
Update the VLLM FIM test to be more predictable

rdimitrov:
The VLLM/Ollama tests have always been too flaky, mainly due to the size of the models we are using there (0.5B). This should hopefully improve this. 

Another thing I noticed was sometimes part of the tests are executed twice for no apparent reason, but I'd address this in a separate PR if I can figure out why it happens.


jhrozek (approved):
thank you!

PR #1304, opened 2025-03-22
Fix Anthropic FIM with muxing.

blkt:
In the context of muxing, the code determining which mapper to use when receiving requests to be routed towards Anthropic was relying in `is_fim_request` only, and was not taking into account if the actual endpoint receiving the request was the legacy one (i.e. `/completions`) or the current one (i.e. `/chat/completions`). This caused the use of the wrong mapper, which led to an empty text content for the FIM request.

A better way to determine which mapper to use is looking at the effective type, since that's the real source of truth for the translation.


blkt:
OK, a re-run of Anthropic integration tests was successful.

jhrozek (approved):

PR #1376, opened 2025-04-28
Remove Ollama from Integration Tests

lukehinds:
Need to test this by pushing


tests/integration/integration_tests.py
```diff
@@ -376,8 +376,13 @@ async def main():
...
             sys.exit(1)
-    else:
-        selected_providers = available_providers
+        else:
+            selected_providers = available_providers
+
+    # Exclude ollama provider if it's in the list
+    if "ollama" in selected_providers:
+        selected_providers.remove("ollama")
+        logger.info("Excluded 'ollama' provider from tests.")
```
rdimitrov, line 385: I think we can leave that there as we might want to run the test locally. 

The change in the workflow yml file should be sufficient to skip the ollama tests from running 👍

blkt (dismissed):

rdimitrov:
@blkt @lukehinds - I see Michelangelo liked the suggestions so I hope you don't mind that I'll push on top of this PR to revert only that change so we can merge the rest

blkt (approved):

